PR #4, opened 2018-07-05
Bradley/flexcpu

powerjg:
Creating this so I can comment on changes.

### DIFF
--- a/src/cpu/flexcpu/generic_reg.hh
+++ b/src/cpu/flexcpu/generic_reg.hh
@@ -1,0 +1,299 @@
+/*
+ * Copyright (c) 2018 The Regents of The University of California
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Authors: Bradley Wang
+ */
+
+#include <type_traits>
+#include <typeinfo>
+
+#include "arch/registers.hh"
+#include "config/the_isa.hh"
+#include "cpu/reg_class.hh"
+
+#ifndef __CPU_FLEXCPU_GENERIC_REG_HH__
+#define __CPU_FLEXCPU_GENERIC_REG_HH__
+
+/**
+ * Generic register value container. Useful for storing results of instructions
+ * or creating unified data structures for holding register values, without
+ * needing to know the type of register at compile-time.
+ */
+class GenericReg {
+  protected:
+    /**
+     * Union storage type, so that GenericReg is a fixed-size class. This
+     * allows simple construction of vectors/arrays/etc. of GenericReg.
+     */
+    union RegVal
+    {
+        TheISA::IntReg intReg;
+        TheISA::FloatReg floatReg;
+        TheISA::FloatRegBits floatRegBits;
+        TheISA::VecRegContainer vecReg;
+        TheISA::VecElem vecElem;
+        TheISA::CCReg ccReg;
+        TheISA::MiscReg miscReg;
+    };
+
+    RegVal _value;
+    RegClass _regClass; // Used for enforcement, as well as identification.
+
+  public:
+    template<typename T>
+    explicit GenericReg(const T& value, const RegClass reg_class):
+        _value({0}),
+        _regClass(reg_class)
+    {
+        switch(_regClass) {
+          case RegClass::IntRegClass:
+            _value.intReg = value;
+            break;
+          case RegClass::FloatRegClass:
+            if (typeid(T) == typeid(TheISA::FloatRegBits)) {
+                _value.floatRegBits = value;
+            } else {
+                _value.floatReg = value;
+            }
+            break;
+          case RegClass::VecRegClass:
+            panic("Templated call should not be used for VecRegClass!");
+            break;
+          case RegClass::VecElemClass:
+            _value.vecElem = value;
+            break;
+          case RegClass::CCRegClass:
+            _value.ccReg = value;
+            break;
+          case RegClass::MiscRegClass:
+            _value.miscReg = value;
+            break;
+          default:
+            panic("Tried to create a GenericReg with unknown type!");
+        }
+    }
+
+    explicit GenericReg(const TheISA::VecRegContainer& value,
+                        const RegClass reg_class):
+        _value({0}),
+        _regClass(reg_class)
+    {
+        assert(reg_class == RegClass::VecRegClass);
+        _regClass = reg_class;
+        _value.vecReg = value;
+    }
+
+    GenericReg(const GenericReg& other):
+        _value({0}),
+        _regClass(other.getRegClass())
+    {
+        switch(_regClass) {
+          case RegClass::IntRegClass:
+            _value.intReg = other.asIntReg();
+            break;
+          case RegClass::FloatRegClass:
+            _value.floatReg = other.asFloatReg();
+            break;
+          case RegClass::VecRegClass:
+            _value.vecReg = other.asVecReg();
+            break;
+          case RegClass::VecElemClass:
+            _value.vecElem = other.asVecElem();
+            break;
+          case RegClass::CCRegClass:
+            _value.ccReg = other.asCCReg();
+            break;
+          case RegClass::MiscRegClass:
+            _value.miscReg = other.asMiscReg();
+            break;
+          default:
+            panic("Tried to copy a GenericReg with unknown type!");
+        }
+    }
+
+    RegClass getRegClass() const
+    {
+        return _regClass;
+    }
+
+    TheISA::IntReg& asIntReg()
+    {
+        assert(getRegClass() == IntRegClass);
+        return _value.intReg;
+    }
+
+    const TheISA::IntReg& asIntReg() const
+    {
+        assert(getRegClass() == IntRegClass);
+        return _value.intReg;
+    }
+
+    TheISA::FloatReg& asFloatReg()
+    {
+        assert(getRegClass() == FloatRegClass);
+        return _value.floatReg;
+    }
+
+    const TheISA::FloatReg& asFloatReg() const
+    {
+        assert(getRegClass() == FloatRegClass);
+        return _value.floatReg;
+    }
+
+    TheISA::FloatRegBits& asFloatRegBits()
+    {
+        assert(getRegClass() == FloatRegClass);
+        return _value.floatRegBits;
+    }
+
+    const TheISA::FloatRegBits& asFloatRegBits() const
+    {
+        assert(getRegClass() == FloatRegClass);
+        return _value.floatRegBits;
+    }
+
+    TheISA::VecRegContainer& asVecReg()
+    {
+        assert(getRegClass() == VecRegClass);
+        return _value.vecReg;
+    }
+
+    const TheISA::VecRegContainer& asVecReg() const
+    {
+        assert(getRegClass() == VecRegClass);
+        return _value.vecReg;
+    }
+
+    TheISA::VecElem& asVecElem()
+    {
+        assert(getRegClass() == VecElemClass);
+        return _value.vecElem;
+    }
+
+    const TheISA::VecElem& asVecElem() const
+    {
+        assert(getRegClass() == VecElemClass);
+        return _value.vecElem;
+    }
+
+    TheISA::CCReg& asCCReg()
+    {
+        assert(getRegClass() == CCRegClass);
+        return _value.ccReg;
+    }
+
+    const TheISA::CCReg& asCCReg() const
+    {
+        assert(getRegClass() == CCRegClass);
+        return _value.ccReg;
+    }
+
+    TheISA::MiscReg& asMiscReg()
+    {
+        assert(getRegClass() == MiscRegClass);
+        return _value.miscReg;
+    }
+
+    const TheISA::MiscReg& asMiscReg() const
+    {
+        assert(getRegClass() == MiscRegClass);
+        return _value.miscReg;
+    }
+
+    GenericReg& operator=(const GenericReg& other)
+    {
+        _regClass = other.getRegClass();
+        switch(_regClass) {
+          case RegClass::IntRegClass:
+            _value.intReg = other.asIntReg();
+            break;
+          case RegClass::FloatRegClass:
+            _value.floatReg = other.asFloatReg();
+            break;
+          case RegClass::VecRegClass:
+            _value.vecReg = other.asVecReg();
+            break;
+          case RegClass::VecElemClass:
+            _value.vecElem = other.asVecElem();
+            break;
+          case RegClass::CCRegClass:
+            _value.ccReg = other.asCCReg();
+            break;
+          case RegClass::MiscRegClass:
+            _value.miscReg = other.asMiscReg();
+            break;
+          default:
+            panic("Tried to copy a GenericReg with unknown type!");
+        }
+        return *this;
+    }
+
+    template<typename T>
+    GenericReg& set(const T& value, const RegClass reg_class)
+    {
+        _regClass = reg_class;
+        switch(_regClass) {
+          case RegClass::IntRegClass:
+            _value.intReg = value;
+            break;
+          case RegClass::FloatRegClass:
+            if (typeid(T) == typeid(TheISA::FloatRegBits)) {
+                _value.floatRegBits = value;
+            } else {
+                _value.floatReg = value;
+            }
+            break;
+          case RegClass::VecRegClass:
+            panic("Templated call should not be used for VecRegClass!");
+            break;
+          case RegClass::VecElemClass:
+            _value.vecElem = value;
+            break;
+          case RegClass::CCRegClass:
+            _value.ccReg = value;
+            break;
+          case RegClass::MiscRegClass:
+            _value.miscReg = value;
+            break;
+          default:
+            panic("Tried to set a GenericReg with unknown type!");
+        }
+        return *this;
+    }
+
+    GenericReg& set(const TheISA::VecRegContainer& value,
+                             const RegClass reg_class)
+    {
+        assert(reg_class == RegClass::VecRegClass);
+        _regClass = reg_class;
+        _value.vecReg = value;
+
+        return *this;
+    }
+};
+
+#endif // __CPU_FLEXCPU_GENERIC_REG_HH__
--- a/src/cpu/flexcpu/inflight_inst.cc
+++ b/src/cpu/flexcpu/inflight_inst.cc
@@ -1,0 +1,648 @@
+/*
+ * Copyright (c) 2018 The Regents of The University of California
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Authors: Bradley Wang
+ */
+
+#include "cpu/flexcpu/inflight_inst.hh"
+
+using namespace std;
+using IntReg = TheISA::IntReg;
+using PCState = TheISA::PCState;
+using FloatReg = TheISA::FloatReg;
+using FloatRegBits = TheISA::FloatRegBits;
+using MiscReg = TheISA::MiscReg;
+
+using CCReg = TheISA::CCReg;
+using VecElem = TheISA::VecElem;
+
+InflightInst::InflightInst(ThreadContext* backing_context,
+                           MemIface* backing_mem_iface,
+                           InstSeqNum seq_num,
+                           const PCState& pc_,
+                           StaticInstPtr inst_ref):
+    backingContext(backing_context),
+    backingMemoryInterface(backing_mem_iface),
+    _status(Waiting),
+    _seqNum(seq_num),
+    _pcState(pc_)
+{
+    staticInst(inst_ref);
+}
+
+void
+InflightInst::addCompletionCallback(std::function<void()> callback)
+{
+    completionCallbacks.push_back(callback);
+}
+
+void
+InflightInst::addDependency(shared_ptr<InflightInst> parent)
+{
+    ++remainingDependencies;
+    parent->addCompletionCallback([this]() {
+        --remainingDependencies;
+
+        if (!remainingDependencies) {
+            notifyReady();
+        }
+    });
+}
+
+void
+InflightInst::addReadyCallback(std::function<void()> callback)
+{
+    readyCallbacks.push_back(callback);
+}
+
+void
+InflightInst::notifyComplete()
+{
+    status(Complete);
+
+    for (function<void()>& callback_func : completionCallbacks) {
+        callback_func();
+    }
+}
+
+void
+InflightInst::notifyReady()
+{
+    for (function<void()>& callback_func : readyCallbacks) {
+        callback_func();
+    }
+}
+
+void
+InflightInst::setDataSource(int8_t src_idx, DataSource source)
+{
+    sources[src_idx] = source;
+}
+
+const StaticInstPtr&
+InflightInst::staticInst(const StaticInstPtr& inst_ref)
+{
+    instRef = inst_ref;
+
+    if (inst_ref) {
+        const int8_t num_dsts = inst_ref->numDestRegs();
+        for (int8_t i = 0; i < num_dsts; ++i) {
+            // TODO consider setting class to corresponding RegId, and filling
+            // values in, in-case of conditional register access.
+            // Fill results table with dummy values
+            results.push_back(GenericReg(0, IntRegClass));
+            resultValid.push_back(false);
+        }
+
+        const int8_t num_srcs = inst_ref->numSrcRegs();
+        for (int8_t i = 0; i < num_srcs; ++i) {
+            sources.push_back({});
+        }
+    }
+
+    return instRef;
+}
+
+// BEGIN ExecContext interface functions
+
+IntReg
+InflightInst::readIntRegOperand(const StaticInst* si, int op_idx)
+{
+    // Sanity check
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isIntReg());
+
+    if (reg_id.isZeroReg()) return 0;
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asIntReg();
+    } else {
+        // Either the producing instruction has already been committed, or no
+        // dependency was in-flight at the time that this instruction was
+        // issued.
+        return backingContext->readIntReg(reg_id.index());
+    }
+}
+
+void
+InflightInst::setIntRegOperand(const StaticInst* si, int dst_idx, IntReg val)
+{
+    const RegId& reg_id = si->destRegIdx(dst_idx);
+    assert(reg_id.isIntReg());
+
+    results[dst_idx].set(reg_id.isZeroReg() ? 0 : val, IntRegClass);
+    resultValid[dst_idx] = true;
+}
+
+FloatReg
+InflightInst::readFloatRegOperand(const StaticInst* si, int op_idx)
+{
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isFloatReg());
+
+    if (reg_id.isZeroReg()) return 0;
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asFloatReg();
+    } else {
+        // Either the producing instruction has already been committed, or no
+        // dependency was in-flight at the time that this instruction was
+        // issued.
+        return backingContext->readFloatReg(reg_id.index());
+    }
+}
+
+FloatRegBits
+InflightInst::readFloatRegOperandBits(const StaticInst* si, int op_idx)
+{
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isFloatReg());
+
+    if (reg_id.isZeroReg()) return 0;
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asFloatRegBits();
+    } else {
+        // Either the producing instruction has already been committed, or no
+        // dependency was in-flight at the time that this instruction was
+        // issued.
+        return backingContext->readFloatRegBits(reg_id.index());
+    }
+}
+
+void
+InflightInst::setFloatRegOperand(const StaticInst* si, int dst_idx,
+                                 FloatReg val)
+{
+    const RegId& reg_id = si->destRegIdx(dst_idx);
+    assert(reg_id.isFloatReg());
+
+    results[dst_idx].set(reg_id.isZeroReg() ? 0 : val, FloatRegClass);
+    resultValid[dst_idx] = true;
+}
+
+void
+InflightInst::setFloatRegOperandBits(const StaticInst* si, int dst_idx,
+                                     FloatRegBits val)
+{
+    const RegId& reg_id = si->destRegIdx(dst_idx);
+    assert(reg_id.isFloatReg());
+
+    results[dst_idx].set(reg_id.isZeroReg() ? 0 : val, FloatRegClass);
+    resultValid[dst_idx] = true;
+}
+
+const TheISA::VecRegContainer&
+InflightInst::readVecRegOperand(const StaticInst* si, int op_idx) const
+{
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isVecReg());
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asVecReg();
+    } else {
+        // Either the producing instruction has already been committed, or no
+        // dependency was in-flight at the time that this instruction was
+        // issued.
+        return backingContext->readVecReg(reg_id);
+    }
+}
+
+TheISA::VecRegContainer&
+InflightInst::getWritableVecRegOperand(const StaticInst* si, int op_idx)
+{
+    // TODO this one may produce bad results, need to rethink this.
+
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isVecReg());
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asVecReg();
+    } else {
+        // Either the producing instruction has already been committed, or no
+        // dependency was in-flight at the time that this instruction was
+        // issued.
+        return backingContext->getWritableVecReg(reg_id);
+    }
+}
+
+void
+InflightInst::setVecRegOperand(const StaticInst* si, int dst_idx,
+                               const TheISA::VecRegContainer& val)
+{
+    const RegId& reg_id = si->destRegIdx(dst_idx);
+    assert(reg_id.isVecReg());
+
+    // NOTE This assignment results in two calls to copy the VecRegContainer
+    // object, one to construct the temporary GenericReg, the other to
+    // perform the copy into the results table. There may be a more efficient
+    // way to do this, perhaps with an in-place "set" function.
+    results[dst_idx].set(val, VecRegClass);
+}
+
+ConstVecLane8
+InflightInst::readVec8BitLaneOperand(const StaticInst* si, int op_idx) const
+{
+    panic("readVec8BitLaneOperand() not implemented!");
+    // To make sure compiler doesn't complain
+    return backingContext->readVec8BitLaneReg(si->srcRegIdx(op_idx));
+    // TODO
+}
+
+ConstVecLane16
+InflightInst::readVec16BitLaneOperand(const StaticInst* si, int op_idx) const
+{
+    panic("readVec16BitLaneOperand() not implemented!");
+    // To make sure compiler doesn't complain
+    return backingContext->readVec16BitLaneReg(si->srcRegIdx(op_idx));
+    // TODO
+}
+
+ConstVecLane32
+InflightInst::readVec32BitLaneOperand(const StaticInst* si, int op_idx) const
+{
+    panic("readVec32BitLaneOperand() not implemented!");
+    // To make sure compiler doesn't complain
+    return backingContext->readVec32BitLaneReg(si->srcRegIdx(op_idx));
+    // TODO
+}
+
+ConstVecLane64
+InflightInst::readVec64BitLaneOperand(const StaticInst* si, int op_idx) const
+{
+    panic("readVec64BitLaneOperand() not implemented!");
+    // To make sure compiler doesn't complain
+    return backingContext->readVec64BitLaneReg(si->srcRegIdx(op_idx));
+    // TODO
+}
+
+
+void
+InflightInst::setVecLaneOperand(const StaticInst* si, int dst_idx,
+                        const LaneData<LaneSize::Byte>& val)
+{
+    panic("setVecLaneOperand() not implemented!");
+    // TODO
+}
+
+void
+InflightInst::setVecLaneOperand(const StaticInst* si, int dst_idx,
+                        const LaneData<LaneSize::TwoByte>& val)
+{
+    panic("setVecLaneOperand() not implemented!");
+    // TODO
+}
+
+void
+InflightInst::setVecLaneOperand(const StaticInst* si, int dst_idx,
+                        const LaneData<LaneSize::FourByte>& val)
+{
+    panic("setVecLaneOperand() not implemented!");
+    // TODO
+}
+
+void
+InflightInst::setVecLaneOperand(const StaticInst* si, int dst_idx,
+                        const LaneData<LaneSize::EightByte>& val)
+{
+    panic("setVecLaneOperand() not implemented!");
+    // TODO
+}
+
+VecElem
+InflightInst::readVecElemOperand(const StaticInst* si, int op_idx) const
+{
+    panic("readVecElemOperand() not implemented!");
+    return 0;
+    // TODO
+}
+
+void
+InflightInst::setVecElemOperand(const StaticInst* si, int dst_idx,
+                                const VecElem val)
+{
+    panic("setVecElemOperand() not implemented!");
+    // TODO
+}
+
+CCReg
+InflightInst::readCCRegOperand(const StaticInst* si, int op_idx)
+{
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isCCReg());
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asCCReg();
+    } else {
+        // Either the producing instruction has already been committed, or no
+        // dependency was in-flight at the time that this instruction was
+        // issued.
+        return backingContext->readCCReg(reg_id.index());
+    }
+}
+
+void
+InflightInst::setCCRegOperand(const StaticInst* si, int dst_idx, CCReg val)
+{
+    const RegId& reg_id = si->destRegIdx(dst_idx);
+    assert(reg_id.isCCReg());
+
+    results[dst_idx].set(val, CCRegClass);
+    resultValid[dst_idx] = true;
+}
+
+MiscReg
+InflightInst::readMiscRegOperand(const StaticInst* si, int op_idx)
+{
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isMiscReg());
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asMiscReg();
+    } else {
+        // Either the producing instruction has already been committed, or no
+        // dependency was in-flight at the time that this instruction was
+        // issued.
+        return backingContext->readMiscReg(reg_id.index());
+    }
+}
+
+void
+InflightInst::setMiscRegOperand(const StaticInst* si, int dst_idx,
+                                const MiscReg& val)
+{
+    const RegId& reg_id = si->destRegIdx(dst_idx);
+    assert(reg_id.isMiscReg());
+
+    results[dst_idx].set(val, MiscRegClass);
+    resultValid[dst_idx] = true;
+}
+
+MiscReg
+InflightInst::readMiscReg(int misc_reg)
+{
+    panic("readMiscReg() not implemented!");
+    return 0;
+    // TODO
+}
+
+void
+InflightInst::setMiscReg(int misc_reg, const MiscReg& val)
+{
+    panic("setMiscReg() not implemented!");
+    // TODO
+}
+
+PCState
+InflightInst::pcState() const
+{
+    return _pcState;
+}
+
+void
+InflightInst::pcState(const PCState& val)
+{
+    _pcState = val;
+}
+
+Fault
+InflightInst::readMem(Addr addr, uint8_t *data, unsigned int size,
+                Request::Flags flags)
+{
+    if (backingMemoryInterface) {
+        return backingMemoryInterface->readMem(shared_from_this(), addr, data,
+                                               size, flags);
+    } else {
+        panic("Attempted to readMem() without a memory interface!");
+        return NoFault;
+    }
+}
+
+Fault
+InflightInst::initiateMemRead(Addr addr, unsigned int size,
+                              Request::Flags flags)
+{
+    if (backingMemoryInterface) {
+        return backingMemoryInterface->initiateMemRead(shared_from_this(),
+                                                       addr, size, flags);
+    } else {
+        panic("Attempted to initiateMemRead() without a memory interface!");
+        return NoFault;
+    }
+}
+
+Fault
+InflightInst::writeMem(uint8_t* data, unsigned int size, Addr addr,
+                       Request::Flags flags, uint64_t* res)
+{
+    if (backingMemoryInterface) {
+        return backingMemoryInterface->writeMem(shared_from_this(), data, size,
+                                                addr, flags, res);
+    } else {
+        panic("Attempted to writeMem() without a memory interface!");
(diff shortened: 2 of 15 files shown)


src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -0,0 +1,158 @@
...
+    assert(tid < numThreads);
+
+    // Get the ThreadContext interface pointer from BaseCPU::getContext(), for
+    // flexibility in how the ThreadContext is retrieved (in case we change
+    // from using SimpleThread).
+    ThreadContext* tc = getContext(tid);
+    if (tc->status() == ThreadContext::Suspended) {
+        tc->activate();
+    }
+}
```
powerjg: No newline. I'll stop commenting on this, but it should be fixed in the rest of the code.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,242 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Authors: Bradley Wang
+ */
+
+#ifndef __SIMPLE_DATAFLOW_CPU_HH__
```
powerjg: This should have the whole path: __CPU_FLEXCPU_SIMPLE_DATAFLOW_CPU_HH__

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,242 @@
...
+    // BEGIN Internal class definitions
+
+    class DataPort : public MasterPort
+    {
+      protected:
+        SimpleDataflowCPU* cpu;
+      public:
+        DataPort(const std::string& name_, SimpleDataflowCPU* cpu_):
+            MasterPort(name_, cpu_),
+            cpu(cpu_)
```
powerjg: Could be on line above.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,242 @@
...
+    };
+
+    // END Internal class definitions
+
+
+    // BEGIN Internal state variables
+
+    DataPort _dataPort;
+    InstPort _instPort;
+    vector<SDCPUThread> threads;
```
powerjg: std::vector

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,242 @@
...
+     *  instruction
+     * @param callback_func The function reference for purposes of handling the
+     *  result of the execution.
+     * 
+     * @return Whether or not this request has been acknowledged. If this
+     *  function returns false, the requester should not expect callback_func
+     *  to be called in the future. May return false as a result of 
+     */
+    bool requestExecution(StaticInst& inst, ExecContext& context,
+        std::function<void(Fault)> callback_func);
```
powerjg: I'd tab this over so it's in line with "StaticInst", if that's not too long.

PS: I like this idea of a callback function. I hope it works out!

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,242 @@
...
+     * @param inst_vaddr The virtual address we want translated
+     * @param callback_func The function reference for purposes of handling the
+     *  result of the translation
+     * 
+     * @return Whether or not this request has been acknowledged. If this
+     *  function returns false, the requester should not expect callback_func
+     *  to be called in the future.
+     */
+    bool requestInstAddrTranslation(Addr inst_vaddr,
+        std::function<void(Fault,BaseTLB::Translation&)> callback_func);
```
powerjg: I'd give this one more space so it's in line with "request". Just personal preference, though.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,242 @@
...
+     *  send through the Port
+     * @param callback_func The function reference for purposes of handling the
+     *  result of the memory request
+     * 
+     * @return Whether or not this request has been acknowledged. If this
+     *  function returns false, the requester should not expect callback_func
+     *  to be called in the future.
+     */
+    bool requestInstruction(Addr inst_paddr,
+        std::function<void(Fault,TheISA::MachInst)> callback_func);
```
powerjg: Again, I'm very excited about this callback function idea. However, my gut says there's a chance that it won't work out.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,71 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Authors: Bradley Wang
+ */
+
+#include "cpu/flexcpu/simple_dataflow_thread.hh"
+
+// Fullsystem mode constructor
+SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
```
powerjg: BTW, no need for all of these silly underscores.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,154 @@
...
+
+    class SDCPUExecContext : public ExecContext
+    {
+      protected:
+        SDCPUThread& backingState;
+        
+      public:
+        SDCPUExecContext(SDCPUThread& backing_state):
+            backingState(backing_state)
+        { }
```
powerjg: Missing a few functions ;)

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,154 @@
...
+     * our dynamic instructions.
+     */
+    typedef uint64_t InstTag;
+    InstTag numInstsCommitted = 0;
+    InstTag numInstsSpeculated = 0;
+
+    /**
+     * The SimpleThread class
+     */
+    SimpleThread _committedState;
```
powerjg: Interesting idea... I hope this works out :)

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,154 @@
...
+
+    // BEGIN Speculative state
+
+    /**
+     * Instruction execution will pass through an ExecContext interface,
+     * therefore in most cases, it will be tied to speculative state, instead
+     * of the "real" state of the system.
+     */
+    SDCPUExecContext _isaInterface;
```
powerjg: I think we talked about needing to have one of these per instruction.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,154 @@
...
+
+    TheISA::PCState _speculativePC;
```
powerjg: We might want to have more than one speculative PC. Maybe not, though. Something to keep in mind.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,154 @@
...
+     * to begin execution of any instructions whose dependencies are all
+     * available.
+     * 
+     * (Note: Assumes in-order issue.)
+     * 
+     * @param inst The instruction to issue.
+     */
+    void issueInstruction(const StaticInst& inst);
+
+    void executeAvailableInstructions();
```
powerjg: I think this function should be in the CPU. The CPU is what will choose from all of the available instructions, which may be from many different threads.

The CPU will look at the "TODO" table you talk about above and execute as many instructions as it can. We don't need to worry about the priority/fairness of this. Now that I'm thinking about it, we obviously should execute the oldest instructions (that are ready) first.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,154 @@
...
+
+    void onBeginFetch();
```
powerjg: These 3 functions make sense. Be sure to doxygen them! ;)

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -53,19 +53,63 @@ SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
...
 void
-SDCPUThreadState::startup()
+SDCPUThread::onBeginFetch()
+{
+    // Retrieve the value of the speculative PC as a raw Addr value.
+    Addr inst_vaddr = _speculativePC.instAddr();
+
+    RequestPtr fetch_req = std::make_shared<Request>();
+
+    fetch_req->taskId(_cpuRef->taskId()); // Necessary? TimingSimpleCPU seems
```
powerjg: This is for tracking statistics. I think it's best to leave it in.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -53,19 +53,63 @@ SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
...
+                                          // to do this...
+    // TODO continue setting up the request before sending it
+
+    if (!_cpuRef->requestInstAddrTranslation(fetch_req,
+        [this](Fault f, RequestPtr& r) {onPCTranslated(f, r);})) {
```
powerjg: I feel less excited about the function pointers now. Will the CPU ever call a different function other than `onPCTranslated` after completing this request? If not, I think it would be clearer to pass in `this` and have the CPU call `this->onPCTranslated`. By using a function pointer here, you're hiding the context of the thread.

However, I'm willing to wait this out and see where it goes. I think it could be interesting for the CPU to be completely agnostic of the thread type, but I'm not sure if this characteristic will be used. If it's not used, then I'm not sure it's worth it.

One final thought, I do like how it, in some sense, keeps the flow control information local. It's obvious from reading this thread code that after you call `requestInstAddrTranslation` next `onPCTranslated` will be called.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -50,19 +53,39 @@ class SDCPUThread
     class SDCPUExecContext : public ExecContext
     {
       protected:
-        SDCPUThread& backingState;
+        InflightInst& backingState;
```
powerjg: May want to rename this variable something like `dynamicInst` or just `inst` and add a comment.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -50,19 +53,39 @@ class SDCPUThread
...
+    class InflightInst
     {
-        StaticInst& instructionRef;
-        // TODO instruction tag
-        // TODO instruction results
+      public:
+        enum Status
+        {
+            Invalid = 0,
+            Waiting, // Can't be executed because dependencies remain
```
powerjg: You also need a `Ready, // all dependencies ready but waiting to execute`

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -50,19 +53,39 @@ class SDCPUThread
...
+        // What is this instruction I'm executing?
+        InstSeqNum tag;
+        StaticInstPtr instRef;
+
+        // How does this instruction influence my system?
+        vector<InstResult> results;
+        vector<TheISA::CCReg> ccResults;
+        Fault fault;
+
+        // TODO deal with macroops?
```
powerjg: I believe that `InflightInst` will be micro-ops (or just ops) and that the macro-ops will be broken down in decode before the `InflightInst` is created. In fact, you may want to add an `assert` or `panic` in the constructor to check this.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -50,19 +53,39 @@ class SDCPUThread
...
+        // TODO deal with memory operations, a la LSQ?
```
powerjg: You'll likely need an effective address and a `MemoryRequest`-like reference.

Though, maybe some of this is more appropriate in the ExecutionContext above. I'm not sure what the difference between that class and this class is.

Either in this class or `ExecContext` there are a few more things you're going to need.

1. Place to hold all of the source data. When this instruction is decoded you'll need to read the architectural registers that aren't busy *at that time*. Similarly, if the architectural register is busy, then you'll need a place for the preceding instruction to write the data when it's done.
2. A place to hold the data you're going to store.
3. A place to hole information for the memory request (like mentioned above).

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -75,56 +98,118 @@ class SDCPUThread
      * execution of an instruction, this should be freely usable as a tag for
      * our dynamic instructions.
      */
-    typedef uint64_t InstTag;
-    InstTag numInstsCommitted = 0;
-    InstTag numInstsSpeculated = 0;
+    InstSeqNum numInstsCommitted = 0;
```
powerjg: Is this the type that other CPUs use? I think this should just be a `uint64_t`.

src/cpu/flexcpu/simple_dataflow_thread.hh
powerjg: I like this idea even more today.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -75,56 +98,118 @@ class SDCPUThread
...
     /**
-     * Instruction execution will pass through an ExecContext interface,
-     * therefore in most cases, it will be tied to speculative state, instead
-     * of the "real" state of the system.
+     * We hold onto the InstSeqNum of the last instruction successfully
+     * committed, for ease of checking if older instruction dependencies have
+     * already been satisfied.
      */
-    SDCPUExecContext _isaInterface;
+    InstSeqNum lastCommittedInstNum;
```
powerjg: I'd like to hear more details on how this will work. :)

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -75,56 +98,118 @@ class SDCPUThread
...
 
+    /**
+     * We hold a "speculative" PC ahead of the "true" committed PC, in order to
+     * track what it is we're fetching next, prior to instruction issue. The
+     * "true" PC is only advanced during instruction commit.
+     */
     TheISA::PCState _speculativePC;
```
powerjg: I think "fetching pc" would be a better term. Or maybe "decoding pc". I think we're going to want to decouple fetch from decode.

For fetch we're working with *addresses*, not a PC. For decode, we're working with a (potentially speculative) PC.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -75,56 +98,118 @@ class SDCPUThread
...
+
+    InstSeqNum numInstsSpeculated = 0;
```
powerjg: Again `uint64_t`

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -75,56 +98,118 @@ class SDCPUThread
...
+
+    /**
+     * We hold a vector of all in-flight instructions, added during issue and
+     * removed after commit. If we always issue in order, this vector should
+     * always be sorted, in the order that the instructions should be committed
+     * so replacing this with a queue might be a better idea hmm.
```
powerjg: Or maybe a map indexed on the inst number because we might need to grab random instructions. A map would still be sorted, which is nice.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -75,56 +98,118 @@ class SDCPUThread
...
+    /**
+     * Utility function for committing the results of a previously in-flight
+     * instruction. This function should update the _committedState using the
+     * in-flight instruction's stored results, increment numInstsCommitted, and
+     * make no further changes to the thread state.
+     */
+    void commitInstruction(shared_ptr<InflightInst> inst);
+
     // TODO make sure this function is capable of holding off on issue in the
     // case where we've only partially read in a macroOp
```
powerjg: To begin with, use RISC-V and don't worry about this. At least, I think you can ignore this with RISC-V

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -0,0 +1,160 @@
...
+#include "cpu/flexcpu/simple_dataflow_cpu.hh"
+
+using namespace std;
+
+SimpleDataflowCPU::SimpleDataflowCPU(SimpleDataflowCPUParams* params):
+    BaseCPU(params),
+    _dataPort(name() + "._dataPort", this),
+    _instPort(name() + "._instPort", this)
+{
+    fatal_if(FullSystem, "FullSystem not implemented for SimpleDataflowCPU");
```
powerjg: Where is `FullSystem` coming from?

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -0,0 +1,160 @@
...
+        thread.startup();
+    }
+    // TODO add code which needs to be run during the first simulation event
+}
+
+void
+SimpleDataflowCPU::switchOut()
+{
+    // TODO flush speculative state.
+    // TODO clear any outstanding memory requests.
```
powerjg: This is called drain, BTW, and there's a special API for it. In this function, you should just assert that this object *is* drained.

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -0,0 +1,160 @@
...
+
+    BaseCPU::switchOut();
+}
+
+void
+SimpleDataflowCPU::takeOverFrom(BaseCPU* cpu)
+{
+    // TODO may want to ensure clean speculative state prior to attempting?
+
+    BaseCPU::takeOverFrom(cpu); // BaseCPU::takeOverFrom() already handles the
```
powerjg: Shouldn't have to do much more here. You may be able to just use the parent function.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,183 @@
...
+
+// Fullsystem mode constructor
+SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
+                    System* system_, BaseTLB* itb_, BaseTLB* dtb_,
+                    TheISA::ISA* isa_, bool use_kernel_stats_ = true):
+    _cpuPtr(cpu_),
+    _committedState(new SimpleThread(cpu_, tid_, system_, itb_, dtb_, isa_,
+                                     use_kernel_stats_))
+{
+    //
```
powerjg: Panic here?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,183 @@
...
+#include "mem/request.hh"
+
+using namespace std;
+
+// Fullsystem mode constructor
+SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
+                    System* system_, BaseTLB* itb_, BaseTLB* dtb_,
+                    TheISA::ISA* isa_, bool use_kernel_stats_ = true):
+    _cpuPtr(cpu_),
+    _committedState(new SimpleThread(cpu_, tid_, system_, itb_, dtb_, isa_,
```
powerjg: I believe there's no need for `new` here.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,248 @@
...
+     * hold onto the state of a basic CPU, demonstrated by its use in many
+     * other CPU models. Until we find a significant need to hold particular
+     * permanent committed state different from a regular CPU in this library,
+     * there is no reason to redefine the wheel.
+     *
+     * The committed state represents the state resulting from the complete
+     * correct execution of all instructions up to the point represented by
+     * numInstsCommitted.
+     */
+    SimpleThread* _committedState;
```
powerjg: Not a pointer?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,183 @@
...
+void
+SDCPUThread::issueInstruction(StaticInstPtr inst, TheISA::PCState cur_pc)
+{
+    const InstSeqNum seq_num = lastCommittedInstNum + inflightInsts.size();
+
+    shared_ptr<InflightInst> dynamic_inst =
+        make_shared<InflightInst>(getThreadContext(), seq_num, inst, cur_pc);
+
+    dynamic_inst->populateDataDependencies(inflightInsts, lastUses);
+    // TODO update lastUses to include the new instruction's uses
```
powerjg: again, I think this would be easier if `lastUses` was it's own type (e.g., dependencyGraph)

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,183 @@
...
+    inflightInsts.push_back(dynamic_inst);
+    // TODO request execution for any instructions that are ready.
```
powerjg: Not sure if this should be here...

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,183 @@
...
+
+void
+SDCPUThread::onInstructionFetched(Fault fault, const TheISA::MachInst inst)
+{
+    if (fault != NoFault) {
+        // TODO handle fault
+    }
+
+    // Capure the PC state at the point of instruction retrieval.
+    TheISA::PCState pc_state = _speculativePC;
```
powerjg: Umm... Not sure if this is right. Couldn't this be some other PC if this fetch took a long time?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,183 @@
...
+    Addr fetched_addr = (pc_state.instAddr() & BaseCPU::PCMask) + fetchOffset;
```
powerjg: Yeah. I don't think this is right.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,183 @@
...
+    } else { // If we still need to fetch more MachInsts.
+        fetchOffset += sizeof(TheISA::MachInst);
+    }
+
+
+    if (true/* Replace with condition*/) {
+        // If we are still running and know where to fetch, we should
+        // immediately send the necessary requests to the CPU to fetch the next
+        // instruction
+        onBeginFetch();
```
powerjg: Hmm... I think we might want to rethink the fetch part of the pipeline. But let's get something working first.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,248 @@
...
+    InstSeqNum lastCommittedInstNum;
+
+    // BEGIN Speculative state
+
+    /**
+     * We hold a "speculative" PC ahead of the "true" committed PC, in order to
+     * track what it is we're fetching next, prior to instruction issue. The
+     * "true" PC is only advanced during instruction commit.
+     */
+    TheISA::PCState _speculativePC;
```
powerjg: Hmm... I'm thinking we might want to separate out fetching and decoding from the threads. 

Again, let's revisit this after the in-order version is working.

src/cpu/flexcpu/simple_dataflow_thread.hh
powerjg: So, where have we come down on this? Should it be a deque?

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: This should be correct until we want to implement out-of-order/speculative pre-fetching.

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -0,0 +1,261 @@
...
+        Executing, // Request for execution sent, but waiting for results
+        Completed // Results have been received, but not yet committed.
+        // Is a separate state for committed instructions that are still held
+        // valuable? May need to consider this to handle squashing memory
+        // requests...
+    };
+
+    struct DataSource
+    {
+        std::weak_ptr<InflightInst> producer = std::weak_ptr<InflightInst>();
```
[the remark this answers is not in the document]
animalvgamer: AFAIK you can only build a weak_ptr from other weak_ptrs, or a shared_ptr. Technically the default constructor used there is a way to call std::weak_ptr<InflightInst>(std::shared_ptr<InflightInst>(nullptr)).

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -249,6 +249,45 @@ SimpleDataflowCPU::requestInstruction(const RequestPtr& req,
     return true;
 }
 
+bool
+SimpleDataflowCPU::requestMemRead(const RequestPtr& req, ThreadContext* tc,
+                                  MemCallback callback_func)
+{
+    // TODO may need to enqueue request if port is busy
```
powerjg: Make sure there isn't an infinite buffer here. For all of the places where there could be queuing make sure that we don't have implicit infinite queues. When testing, you should test with queue sizes of 1 to bring out bugs.

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -249,6 +249,45 @@ SimpleDataflowCPU::requestInstruction(const RequestPtr& req,
...
+    PacketPtr pkt = Packet::createRead(req);
+
+    uint8_t* mailbox = new uint8_t[req->getSize()]
+    pkt->dataStatic<uint8_t>(mailbox);
```
powerjg: What is this mailbox? It should at least have a comment, but I have a feeling there's a better way to accomplish what you're trying to do here.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -204,11 +208,57 @@ SDCPUThread::onBeginFetch()
...
+void
+SDCPUThread::onDataAddrTranslated(Fault fault, const RequestPtr& req,
+                                  InflightInst& inst, bool write)
+{
+    if (fault != NoFault) {
+        // TODO handle fault.
+        panic("SDCPU not able to handle data translation faults yet!");
+    }
+
+    // TODO set the effective address
```
powerjg: BTW, usually effective address is considered the virtual address of the access and is computed from the register+immediate offset. This is the physical address (usually).

You should be able to do memory disambiguation on the *effective address*. You shouldn't have to use the physical address unless switching address spaces is not serializing (it is).

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -204,11 +208,57 @@ SDCPUThread::onBeginFetch()
...
+    }
+}
+
+void
+SDCPUThread::onExecutionCompleted(Fault fault, InflightInst& inst)
+{
+    // The InflightInst will also match any of its dependencies automatically,
+    // and perform any ready callbacks if any instructions are made ready as a
+    // result of this instruction completed
+    inst.notifyComplete();
```
powerjg: I think you need to call static_inst->completeAcc()?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -223,7 +273,8 @@ SDCPUThread::onPCTranslated(Fault fault, const RequestPtr& req)
     if (!_cpuPtr->requestInstruction(req,
         [this](TheISA::MachInst i) {
             onInstructionFetched(i);
-        })) {
+        })
+    ) {
         panic("The CPU rejected my fetch request and I haven't been programmed"
```
powerjg: The above is really hard to read. It would be better to use a variable for the callback.

```
auto callback = [this](TheISA::MachInst i) { onInstructionFetched(i); };
if (!_cpuPtr->requestInstruction(req, callback)) {
    panic("...");
}
```

In fact, it may be a good idea to do this in many places. I would say it's a good idea if there's anywhere that you use the callback inside of another statement like this.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -299,3 +350,72 @@ SDCPUThread::startup()
...
+    sdCPUThread._cpuPtr->requestDataAddrTranslation(req, tc,
+        [&sdCPUThread, &ref](Fault fault, const RequestPtr& req) {
+            sdCPUThread.onDataAddrTranslated(fault, req, ref, false);
+        });
+
+    return NoFault;
+}
+
+Fault
+SDCPUThread::MemIface::writeMem(InflightInst& ref, uint8_t *data,
```
powerjg: There's a lot of duplicated code between this function and `initiateMemRead()`. Any way to factor some out into its own function?

src/cpu/flexcpu/simple_dataflow_cpu.cc
animalvgamer: It's basically how TimingSimpleCPU does it. It allocates some memory with new in order to have something to pass to the Packet for dataDynamic/Static, then deletes it afterward. Tbh since the lifespan of a packet is well-defined, I actually think that it's the cleanest way to manage it. I guess this was why I was asking if there was a guaranteed recvTimingResp call for every sendTimingReq sent successfully.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: But you said synonyms were a problem? This was just me noting that I had to take care of it at some point (TODO).

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: I was intending for the CPU to make that call prior to the onExecutionCompleted call, since I'm basically letting the CPU distinguish between memory request writebacks and regular writebacks.

In this code, I'm treating both StaticInst::execute() and StaticInst::completeAcc() as some ISA-defined set of tasks that will produce accesses to the ExecContext, after which the instruction is considered Complete.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: Good idea, noted.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: I had that thought too, but I'm also not quite done with the writeMem function yet, since I need to find a good way to pass the uint8_t* data param through to when the actual packet is being sent. Once that is sorted out, I'll think about consolidating the code.

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -31,30 +31,46 @@
 #include "cpu/flexcpu/inflight_inst.hh"
 
 using namespace std;
```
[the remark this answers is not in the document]
animalvgamer: Oh, why's that? I feel like that's a little extreme ._.

src/cpu/flexcpu/generic_reg.hh
```diff
@@ -0,0 +1,299 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Authors: Bradley Wang
+ */
+
+#include <type_traits>
```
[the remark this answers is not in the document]
animalvgamer: It's used on line 77, for typeid()

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -0,0 +1,648 @@
...
+MiscReg
+InflightInst::readMiscRegOperand(const StaticInst* si, int op_idx)
+{
+    const RegId& reg_id = si->srcRegIdx(op_idx);
+    assert(reg_id.isMiscReg());
+
+    const DataSource& source = sources[op_idx];
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
```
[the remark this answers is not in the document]
animalvgamer: I think that it should be up to the thread to enforce that. I implemented it the way I did more to generalize the use of "operands" within the StaticInst interface. Source operands are selected from sources[], and destination operands are held by the InflightInst results[] until the user of InflightInst has a place to put them

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -0,0 +1,648 @@
...
+
+Fault
+InflightInst::initiateMemRead(Addr addr, unsigned int size,
+                              Request::Flags flags)
+{
+    if (backingMemoryInterface) {
+        return backingMemoryInterface->initiateMemRead(shared_from_this(),
+                                                       addr, size, flags);
+    } else {
+        panic("Attempted to initiateMemRead() without a memory interface!");
```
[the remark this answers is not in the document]
animalvgamer: The constructor doesn't stop you from passing nullptr though, and I don't think it should stop you from passing nullptr, since it would be totally reasonable to construct a non-memref InflightInst and not to bother giving it an implementation of this interface.

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -0,0 +1,648 @@
...
+                           MemIface* backing_mem_iface,
+                           InstSeqNum seq_num,
+                           const PCState& pc_,
+                           StaticInstPtr inst_ref):
+    backingContext(backing_context),
+    backingMemoryInterface(backing_mem_iface),
+    _status(Waiting),
+    _seqNum(seq_num),
+    _pcState(pc_)
+{
```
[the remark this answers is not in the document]
animalvgamer: I don't think it should stop you from passing nullptr, since it would be totally reasonable to construct a non-memref InflightInst and not to bother giving it an implementation of this interface.

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -0,0 +1,648 @@
...
+    const shared_ptr<InflightInst> producer = source.producer.lock();
+
+    if (producer) {
+        // If the producing instruction is still in the buffer, grab the result
+        // assuming that it has been produced, and our index is in bounds.
+        assert(producer->isComplete());
+        assert(source.resultIdx >= 0
+            && source.resultIdx < producer->results.size()
+            && producer->resultValid[source.resultIdx]);
+        return producer->getResult(source.resultIdx).asCCReg();
```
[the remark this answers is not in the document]
animalvgamer: Definitely will keep a note on that there, but I don't know how simple it would be to capture specific bit reads/writes without doing the preexecution as we described before...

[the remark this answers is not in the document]
animalvgamer: And even then I'm not sure if it's simple to identify exactly which bits are being written by the instruction, if we want to rename based on bits.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+
+    /**
+     * For use when we need to fetch multiple MachInsts for one decode.
+     */
+    Addr fetchOffset = 0;
+
+    /**
+     * Number of fetch actions we have begun and not finished at this time
+     */
+    size_t outstandingFetches = 0;
```
powerjg: Why needed? It's not used in the code right now. If this is for constraining the fetch bandwidth, I think that should be done by the CPU, not the thread.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+     * have not been committed yet, and we want to keep committed and
+     * speculative state separate.
+     */
+    TheISA::Decoder decoder;
+
+    /**
+     * We hold a list of all in-flight instructions, added during issue and
+     * removed after commit. If we always issue in order, this list should
+     * always be sorted, in the order that the instructions should be committed
+     * so replacing this with a queue might be a better idea hmm.
```
powerjg: Just do it! Use a queue :).

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+     */
+    std::list<std::shared_ptr<InflightInst>> inflightInsts;
+
+    /**
+     * For quick dependency extraction, we keep track of the most recent
+     * write accesses to each unique register.
+     */
+    std::unordered_map<RegId, InflightInst::DataSource> lastUses;
```
powerjg: Cool!

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+
+    InstSeqNum earliestOutstandingFaultingInstNum = 0;
```
powerjg: Unused?

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+     * related events, such as onBeginFetch to kick off any upcoming tasks
+     * necessary to complete that instruction.
+     */
+    void advanceInst();
+
+    /**
+     * Utility function for iterating through the in-flight instruction table
+     * and committing all instructions at once (removing them from the table).
+     */
+    void commitAllAvailableInstructions();
```
powerjg: Like a previous comment. I don't think we want to call this "all"

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+    /**
+     * This function serves as the event handler for when the CPU has completed
+     * the translation we requested for an instruction that does memory access.
+     *
+     * This is an extra step which will only be used for memory accessing
+     * instructions.
+     *
+     * This function should be called after a call to executeInstruction() for
+     * a corresponding memory accessing InflightInst, which should make a call
+     * to TODO on the CPU.
```
powerjg: TODO?

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+     *
+     * @param fault If the translation request generated a fault, it should be
+     *  returned through this parameter.
+     * @param req A reference to a request object which should hold the
+     *  translated physical address.
+     */
+    void onPCTranslated(std::weak_ptr<InflightInst> inst, Fault fault,
+                        const RequestPtr& req);
+
+    bool outstandingFault()
```
powerjg: Not used?

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+     * in order to streamline dependency tracking. Like a scoreboard.
+     *
+     * @param inst The instruction for which we want to track uses.
+     */
+    void populateUses(std::shared_ptr<InflightInst> inst);
+
+    /**
+     * This function checks if the thread should be active at this moment, and
+     * whether or not we have a reasonable PC value ready to be able to
+     * initiate another fetch at this time.
```
powerjg: Could you explain why you need this function (i.e., to deal with control flow instructions)

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+                     bool use_kernel_stats_ = true);
+
+    // Non-fullsystem constructor
+    SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_, System* system_,
+                     Process* process_, BaseTLB* itb_, BaseTLB* dtb_,
+                     TheISA::ISA* isa_);
+
+    // May need to define move constructor, due to how SimpleThread is defined,
+    // if we want to hold instances of these in a vector instead of pointers
+    // gotten from new.
```
powerjg: Ok. I see why you have to have pointers in the vector now. Ignore those comments.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+    };
+
+    // END SDCPUThread Internal definitions
+
+    // BEGIN SDCPUThread Internal variables
+
+    /**
+     * We hold a pointer to the CPU which owns this logical thread.
+     */
+    SimpleDataflowCPU* _cpuPtr;
```
powerjg: Can't we use the `_committedState->getCPUPtr()`? There's no need to replicate this member.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+    Status status() const override { return _committedState->status(); }
+
+    void setStatus(Status new_status) override
+    { _committedState->setStatus(new_status); }
+
+    /// Set the status to Active.
+    void activate() override;
+
+    /// Set the status to Suspended.
+    void suspend() override { _committedState->suspend(); }
```
powerjg: Probably want to `panic` here. Suspending isn't going to work if there are instructions in the pipeline.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+
+    /// Set the status to Halted.
+    void halt() override { _committedState->halt(); }
```
powerjg: Ditto above. Probably panic.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+
+    void dumpFuncProfile() override { _committedState->dumpFuncProfile(); }
+
+    void takeOverFrom(ThreadContext *oldContext) override
+    { _committedState->takeOverFrom(oldContext); }
```
powerjg: Assert that there's no instructions in flight, maybe?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+        shared_ptr<InflightInst> dynamic_inst_ptr =
+            make_shared<InflightInst>(getThreadContext(), &memIface, seq_num,
+                                      next_pc, static_inst);
+
+        inflightInsts.push_back(dynamic_inst_ptr);
+
+        issueInstruction(dynamic_inst_ptr);
+
+        return;
+    }
```
powerjg: else if?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
```
powerjg: else?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+            handleFault(inst_ptr);
+
+            return;
+        }
+
+        if (!inst_ptr->isSquashed()){
+            commitInstruction(inst_ptr);
+        }
+
+        inflightInsts.pop_front();
```
powerjg: For a squashed instruction, is there any cleanup you need to worry about? I guess not since you're passing around weak pointers everywhere and the instruction will be deleted. Could you make a note of that here?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+    }
+}
+
+void
+SDCPUThread::commitInstruction(std::shared_ptr<InflightInst> inst_ptr)
+{
+    assert(inst_ptr->isComplete());
```
powerjg: Do you need to add tracing stuff to this function?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+
+    panic_if(!inst_ptr->isReady(), "Tried to execute an instruction which was "
+                                   "not ready to execute!");
+    const StaticInstPtr static_inst = inst_ptr->staticInst();
+
+    DPRINTF(SDCPUInstEvent, "Beginning executing instruction (seq %d)\n",
+                            inst_ptr->seqNum());
+
+    if (static_inst->isMemRef()) {
+        static_inst->initiateAcc(inst_ptr.get(), nullptr);
```
powerjg: Need to deal with the trace stuff. This will make your life easier later.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+    inst_ptr->fault(fault);
+
+    while (inflightInsts.back() != inst_ptr) {
+        inflightInsts.back()->squash();
+        inflightInsts.pop_back();
+    }
+
+    inst_ptr->squash();
+
+    commitAllAvailableInstructions();
```
powerjg: Hmm... this confuses me. Could you explain more what this is doing here?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+                static_pointer_cast<ExecContext>(inst_ptr), data, callback)) {
+            panic("The CPU rejected my mem access request and I haven't been "
+                  "programmed to know how to continue!!!");
+        }
+    } else { // read
+        if (!_cpuPtr->requestMemRead(req, getThreadContext(),
+                inst_ptr->staticInst(),
+                static_pointer_cast<ExecContext>(inst_ptr),
+                callback)) {
+            panic("The CPU rejected my mem access request and I haven't been "
```
powerjg: We need to think about how to handle this.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+    const StaticInstPtr static_inst = inst_ptr->staticInst();
+
+    // BEGIN Sequential Consistency
+
+    // TODO IMPORTANT this still doesn't make it safe to speculate on memory
+    //      instructions, since if we produce a fault and need to squash, we
+    //      still need a way to inform the cpu to undo stores being sent to
+    //      memory.
+
+    if (static_inst->isMemRef() || static_inst->isMemBarrier()) {
```
powerjg: Add a comment here that you're searching for the youngest memory instruction that has been issued and creating a dependency.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+    const int8_t num_srcs = static_inst->numSrcRegs();
+    for (int8_t src_idx = 0; src_idx < num_srcs; ++src_idx) {
+        const RegId& src_reg = static_inst->srcRegIdx(src_idx);
+        if (src_reg.isZeroReg()) continue;
+
+        const InflightInst::DataSource& last_use = lastUses[src_reg];
+        const shared_ptr<InflightInst> producer = last_use.producer.lock();
+
+        // Attach the dependency if the instruction is still around
+        if (producer && !(producer->isComplete() || producer->isSquashed()))
```
powerjg: { }

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,890 @@
...
+        panic("The CPU rejected my dtb request and I haven't been programmed"
+              " to know how to continue!!!");
+        // TODO request later
+    }
+
+    return NoFault;
+}
+
+Fault
+SDCPUThread::MemIface::writeMem(shared_ptr<InflightInst> inst_ptr,
```
powerjg: There's a lot of  duplicated code between read and write. (This is true in most of the functions that relate to these two things, though this is the worst so far.) Is there any way to factor some of this out?

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+    class DataTransHandler : public BaseTLB::Translation
+    {
+      protected:
+        TranslationCallback callback;
+        bool selfDestruct;
+      public:
+        DataTransHandler(TranslationCallback callback,
+                         bool self_destruct = false):
+            callback(callback),
+            selfDestruct(self_destruct)
```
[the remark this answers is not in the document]
animalvgamer: Well I suppose if I switch away from using new to allocate one :l I just felt like self deletion shouldn't be a default behavior, and I wanted that parameter to be optional I guess

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+            callback(fault, req);
+
+            // It requires assumptions that users of this will be well-behaved
+            // to call "delete this;" in a C++ class, but since this is an
+            // internal class, we should be able to contain access to this line
+            if (selfDestruct) delete this;
+        }
+    };
+
+    class InstTransHandler : public BaseTLB::Translation
```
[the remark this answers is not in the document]
animalvgamer: Yea I think I noticed this too at some point, I probably should refactor to just callbacktranshandler or something

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: So the faulted instruction should be squashed, since the fault meant it couldn't make it safely to completion, and then we make the call to commitAllAvailableInstructions to basically check if inst_ptr is at the head of the inflightInsts table. commitAllAvailableInstructions() also handles faults (at commit time) so since that functionality is shared, I call the same instruction as when I say complete execution, as both events potentially produce commit-time jobs to do.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+    std::list<FetchReq> fetchReqs;
+    std::list<ExecutionReq> executionReqs;
+    std::list<MemAccessReq> memReqs;
+
+    // Note: I was informed that the pointer you get back is the same pointer
+    //       you send through the port, so it's safe to do this mapping. It may
+    //       be safer to do this mapping through the SenderState struct instead
+    //       based on what has been learned since the writing of this though.
+    std::unordered_map<PacketPtr, FetchCallback> outstandingFetches;
+    std::unordered_map<PacketPtr, MemAccessReq> outstandingMemReqs;
```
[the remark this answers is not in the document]
animalvgamer: I mean this was also possible through senderState maybe but I just needed the functionality to know where a call to recvTimingResponse should go

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+    Tick executionTime = clockPeriod();
+
+    // END Internal state variables
+
+
+    // BEGIN Internal functions
+
+    void attemptAllFetchReqs();
+
+    void attemptAllMemReqs();
```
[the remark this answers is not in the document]
animalvgamer: Well it's limited right now by the amount the port can accept at once.

I like the "attempt" terminology here since my intention is to simply try if possible to send them all through the port. It may also send none if the port is busy, so I don't really think a hard sendX scheme makes sense here.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+    // like.
+    std::vector<std::unique_ptr<SDCPUThread>> threads;
+
+    // For use in enqueuing requests if we want to limit number of requests
+    // servicable at once in the future. TODO for now.
+    std::list<TranslationReq> dataTranslationReqs;
+    std::list<TranslationReq> instTranslationReqs;
+    std::list<FetchReq> fetchReqs;
+    std::list<ExecutionReq> executionReqs;
+    std::list<MemAccessReq> memReqs;
```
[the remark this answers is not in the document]
animalvgamer: Oh yea we definitely don't need all of those haha that was just me writing things out when I was definining those request structs, and forgetting to clear them out I guess.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+
+    /**
+     * Event-driven means for classes to request a read access to memory. Upon
+     * the response event from the memory port, the result of completeAcc()
+     * will be passed to the callback function.
+     *
+     * @param req A memory request object, which should contain a completed
+     *  translation and therefore ready to send to memory.
+     * @param callback_func The function reference for purposes of handling the
+     *  result of the memory read.
```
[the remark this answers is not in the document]
animalvgamer: Yea I guess I missed some stuff when refactoring

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+     *
+     * @return Whether or not this request has been acknowledged. If this
+     *  function returns false, the requester should not expect callback_func
+     *  to be called in the future.
+     */
+    bool requestSplitMemRead(const RequestPtr& main, const RequestPtr& low,
```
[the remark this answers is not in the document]
animalvgamer: Mostly because if it's across a cache line, I'm imagining it could also cross a page boundary, meaning that I'd need to track two physical addresses for potential through-memory dependencies. Since the thread is already aware, it was cleaner to just pass the two translated RequestPtrs through from the thread's side.

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -0,0 +1,630 @@
...
+            return; // Stop trying to send requests until we're called again
+                    // in recvReqRetry().
+        }
+
+        DPRINTF(SDCPUCoreEvent, "Sent a fetch request(pa: %#x)\n",
+                                request.packet->req->getPaddr());
+
+        // We mark the control flow so that we know what to call when memory
+        // responds
+        outstandingFetches.emplace(request.packet, request.callback);
```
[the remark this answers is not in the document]
animalvgamer: Yea right now I only ever have one outstanding at a time, but it felt like it made sense to explicitly define a relationship between the packet I sent through the port and the corresponding callback. It means less potential changes if we start doing multiple/pre-fetching.

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -0,0 +1,630 @@
...
+bool
+SimpleDataflowCPU::requestExecution(StaticInstPtr inst,
+                                    weak_ptr<ExecContext> context,
+                                    ExecCallback callback_func)
+{
+    DPRINTF(SDCPUCoreEvent, "requestExecution()\n");
+
+    // Right now we're not limiting execution units, so we just execute all
+    // instructions as soon as possible
+    beginExecution({inst, context, callback_func});
```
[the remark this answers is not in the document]
animalvgamer: When we move to limiting execution units by type, as well as having different execution types take different amounts of time, both functions will get slightly more complex, and having the requestExecution choose either to enqueue or beginExecution, then begin to figure out timing made more sense.

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -0,0 +1,630 @@
...
+SimpleDataflowCPU::InstPort::recvTimingResp(PacketPtr pkt)
+{
+    DPRINTF(SDCPUCoreEvent, "Received _instPort response (pa: %#x)\n",
+                            pkt->req->getPaddr());
+
+    panic_if(!cpu->outstandingFetches.count(pkt),
+        "Received an instruction port response we don't remember sending!");
+
+    const FetchCallback& callback = cpu->outstandingFetches[pkt];
+    MachInst data = *(pkt->getPtr<MachInst>());
```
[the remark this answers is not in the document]
animalvgamer: Wait how is it strange

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -0,0 +1,620 @@
...
+        Fault writeMem(std::shared_ptr<InflightInst> inst, uint8_t *data,
+                       unsigned int size, Addr addr, Request::Flags flags,
+                       uint64_t *res) override;
+    };
+
+    struct SplitRequest {
+        RequestPtr main = nullptr;
+        RequestPtr low = nullptr;
+        RequestPtr high = nullptr;
+    };
```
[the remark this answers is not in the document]
animalvgamer: I don't think so, and here's why:

- The thread needs to know the physical address of all requests made, for tracking data dependencies through memory
- Split requests have two physical addresses, so the thread needs to hold enough information to complete both translations
- The main RequestPtr which will be owned by the PacketPtr passed to completeAcc() needs to be kept alive until that call can be made, otherwise it would have to be rebuilt, which I don't think is a better approach.

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -0,0 +1,261 @@
...
+    // TODO deal with memory operations, a la LSQ?
+
+  public:
+    InflightInst(ThreadContext* backing_context, InstSeqNum seq_num,
+                    StaticInstPtr inst_ref, const TheISA::PCState& next_pc);
+
+    void addCompletionCallback(std::function<void()> callback);
+    void addDependency(std::shared_ptr<InflightInst> parent);
+
+    void addReadyCallback(std::function<void()> callback);
```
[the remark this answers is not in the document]
animalvgamer: So they aren't known at construction time, since we may be filling in dependencies at decode time, or even at memory translation time.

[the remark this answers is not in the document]
animalvgamer: I tried to model these interface functions in the same way Java does its listeners for event-driven UI stuff.

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -0,0 +1,261 @@
...
+     */
+    TheISA::PCState _pcState;
+
+    // Who depends on me?
+    std::vector<std::function<void()>> completionCallbacks;
+
+    // How many dependencies have to be resolved before I can run?
+    size_t remainingDependencies = 0;
+
+    std::vector<std::function<void()>> readyCallbacks;
```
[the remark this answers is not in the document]
animalvgamer: That's right, I should add comments tho

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -0,0 +1,261 @@
...
+
+    void populateDataDependencies(
+        const std::vector<InflightInst>& prior_inflight,
+        const std::unordered_map<RegId, DataSource>& last_uses);
+
+    Status status() const
+    { return _status; }
+
+    Status status(Status status)
+    { return _status = status; }
```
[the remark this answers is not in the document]
animalvgamer: Really? I specifically did it this way because ExecContext had stuff in this fashion

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -0,0 +1,597 @@
...
+    instRef(inst_ref),
+    pcState(next_pc),
+    _status(Waiting)
+{
+    const int8_t num_dsts = inst_ref->numDestRegs();
+    for (int8_t i = 0; i < num_dsts; ++i) {
+        // TODO consider setting class to corresponding RegId, and filling
+        // values in, in-case of conditional register access.
+        // Fill results table with dummy values
+        results.push_back(GenericReg<TheISA::IntReg>(0, IntRegClass));
```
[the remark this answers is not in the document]
animalvgamer: I guess I could define a dummy constructor? Since I'm using the RegClass enum though it would probably still have to have a default value...

src/cpu/flexcpu/simple_dataflow_thread.hh
animalvgamer: So I tried that, but SimpleThread didn't seem to like it when I first did it... I could mess with it again but that's what had been working so far.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+        SimpleDataflowCPU* cpu;
+      public:
+        bool receivedPushback = false;
+
+        DataPort(const std::string& name_, SimpleDataflowCPU* cpu_):
+            MasterPort(name_, cpu_),
+            cpu(cpu_)
+        { }
+
+        virtual ~DataPort()
```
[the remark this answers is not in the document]
animalvgamer: Excuse me?

src/cpu/flexcpu/generic_reg.hh
```diff
@@ -0,0 +1,299 @@
+/*
```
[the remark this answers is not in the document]
animalvgamer: I'll try to figure that out

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -0,0 +1,496 @@
...
+        SimpleDataflowCPU* cpu;
+      public:
+        bool receivedPushback = false;
+
+        InstPort(const std::string& name_, SimpleDataflowCPU* cpu_):
+            MasterPort(name_, cpu_),
+            cpu(cpu_)
+        { }
+
+        virtual ~InstPort()
```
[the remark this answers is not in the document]
animalvgamer: If we're talking about the destructor, I often default to declaring a virtual destructor when defining classes, simply because I rarely feel like there's a reason for a destructor not to be virtual. Maybe it's because I'm used to Java where there aren't restrictions on what classes you can subclass.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: I agree. Right now, the true-false accept-reject nature of the relationship between CPU and thread is kind of in a weird place, because the responsibility of buffering requests or stalling is not necessarily well-defined in this model yet.

PR #6, opened 2018-07-20
Reg class changes (including hash function)

animalvgamer:
Just making a pull request for review of the hash functions.


powerjg:
The commit messages need to be updated.

I believe the first commit (cpu: Removed...) is too long. The first line should be <65 chars IIRC.
(Note: All other lines should be <75 chars. I'm not sure if I'm following that below because it's hard to count in this interface.)

Both need more details.

For example:
cpu: Remove unnecessary file reg_class_impl.hh

Previously, reg_class_impl.hh was required to remove a cyclic dependency in header files
(See http://reviews.gem5.org/r/3754/).
This is no longer required. I tested on various gcc and clang compilers and on MacOS and Linux.

For the other one:
cpu: Add hash function for RegId class

With the hash function, you can now use RegId in an unordered_map

powerjg:
Closing since accepted upstream.

PR #7, opened 2018-07-26
Replace == version with >= version

animalvgamer:
For forward compatibility with later compilers, we should have a greater
than comparison instead of an explicit equality with a particular C++
version.


powerjg:
Closing since it's accepted upstream.

PR #8, opened 2018-07-27
Rebased Testing Library

spwilson2:
This branch is the staging area for the rebased patchset into upstream.


spwilson2:
Use #5 for review comments.

PR #12, opened 2018-08-20
[fixup] Fix bug in split writes in SimpleDataflowCPU

powerjg:
Change-Id: I5d5e05fa87bad7586af793b44311955517b89724

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_cpu.cc
+++ b/src/cpu/flexcpu/simple_dataflow_cpu.cc
@@ -193,6 +193,8 @@ SimpleDataflowCPU::completeMemAccess(const MemAccessReq& req)
             return; // We need to wait for the other to complete as well
 
         pkt = req.split->main;
+        assert(pkt);
+        pkt->makeResponse();
     } else {
         pkt = req.packet;
     }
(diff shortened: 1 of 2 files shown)


powerjg:
It's definitely broken right now... I'll think about it for a few minutes.

powerjg:
How does it look now?

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -481,16 +483,18 @@ SimpleDataflowCPU::requestSplitMemWrite(const RequestPtr& main,
 
     SplitAccCtrlBlk* split_acc = new SplitAccCtrlBlk;
 
+    // This pointer may be used in completeAcc. It probably doesn't need data.
+    // But, we give it data, just to be safe.
```
powerjg: Lol. That was, a lot, of commas, there.

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -193,6 +193,8 @@ SimpleDataflowCPU::completeMemAccess(const MemAccessReq& req)
             return; // We need to wait for the other to complete as well
```
powerjg: In combination with the short sentence before it was silly to read.

PR #15, opened 2018-08-21
cpu: Add explicit serialization logic to flexcpu

animalvgamer:
Change-Id: I408fda8a30d3089f7e329e3a64d108b929166b52

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -630,7 +630,26 @@ SDCPUThread::populateDependencies(shared_ptr<InflightInst> inst_ptr)
     //         inst->addDependency(last_inst);
     // }
 
+    /*
+    * Note: An assumption is being made that inst_ptr is at the end of the
+    *       inflightInsts buffer.
+    */
+
     const StaticInstPtr static_inst = inst_ptr->staticInst();
+
+    // BEGIN ISA explicit serialization
+
+    const bool we_ser_before = static_inst->isSerializeBefore();
+    for (auto& prior : inflightInsts) {
+        if (prior == inst_ptr || prior->isComplete() || prior->isSquashed()) {
+            continue;
+        }
+
+        const bool they_ser_after = prior->staticInst()->isSerializeAfter();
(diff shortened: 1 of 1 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -630,8 +630,27 @@ SDCPUThread::populateDependencies(shared_ptr<InflightInst> inst_ptr)
...
+    // BEGIN ISA explicit serialization
+
+    const bool we_ser_before = static_inst->isSerializeBefore();
+    for (auto& prior : inflightInsts) {
+        if (prior == inst_ptr || prior->isComplete() || prior->isSquashed()) {
+            continue;
+        }
+
+        const bool they_ser_after = prior->staticInst()->isSerializeAfter();
+        if (we_ser_before || they_ser_after) inst_ptr->addDependency(prior);
```
powerjg: Can you add a dependency just on the (one) previous instruction?

No... I guess this makes sense. 

One way you *could* do it would be to add a "commit dependency" and then just say the previous instruction has to commit. However, that might be stronger than you need.

Here's a question: can you execute a serializing instruction before older instructions commit (but they have executed)? If so, this logic works. If not, you might need to add a different kind of call back for commit time.

animalvgamer: Yea that's a fair point... gem5 isn't very specific about what the flags on the StaticInst mean though, so I'm not sure how to answer that...

powerjg: I'm a little worried about miscreg changes. Can we add an assert on miscreg changes to make sure they only affect architectural state when their inflight_inst is at the head of the "ROB" (i.e., committing next)? This can be another patch.

animalvgamer: What do you mean by that? Like no state is updated except at commit time so...

powerjg: The miscreg isn't renamed, right? So when an instruction writes the miscreg via the exec context it directly updates the thread state.

Oh... I see that in the code this will generate a panic :). This is something to think about when implementing the read/write miscreg.

PR #16, opened 2018-08-21
cpu: Cleanup flexcpu style and improved debug messages

animalvgamer:
Change-Id: I7ae4d2a3993a7c5bae0ec5d0969b56f31c9feb60

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -100,9 +100,9 @@ SDCPUThread::activate()
 void
 SDCPUThread::advanceInst()
 {
-    DPRINTF(SDCPUThreadEvent, "advanceInst()\n");
-
     const InstSeqNum seq_num = lastCommittedInstNum + inflightInsts.size() + 1;
+
+    DPRINTF(SDCPUThreadEvent, "advanceInst(seq %d)\n", seq_num);
 
     TheISA::PCState next_pc = getNextPC();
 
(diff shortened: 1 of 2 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -186,10 +186,10 @@ SDCPUThread::commitInstruction(std::shared_ptr<InflightInst> inst_ptr)
 {
     assert(inst_ptr->isComplete());
 
-    StaticInstPtr static_inst = inst_ptr->staticInst();
-
     DPRINTF(SDCPUInstEvent,
-             "Committing instruction (seq %d)\n", inst_ptr->seqNum());
+            "Committing instruction (seq %d)\n", inst_ptr->seqNum());
+
+    StaticInstPtr static_inst = inst_ptr->staticInst();
```
powerjg: ? Is this needed?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -505,6 +503,9 @@ SDCPUThread::onExecutionCompleted(weak_ptr<InflightInst> inst, Fault fault)
         return;
     }
 
+    // if (!inst->readPredicate()) // TODO ensure this is enforced.
```
powerjg: If this isn't needed, it should be deleted.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: Well it was more to line up the parameter to DPRINTF, but I moved the call to inst_ptr->staticInst() later to right before I needed the variable I guess

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: Oh that was just a commented line I added because I noticed either O3 or TimingSimple doing that so I was like maybe I should take a note

powerjg: Make sense.

src/cpu/flexcpu/simple_dataflow_thread.cc
powerjg: Oh I see. Ok.

PR #17, opened 2018-08-21
Actually fixes build errors on .fast

powerjg:
Change-Id: Ie91d645dbf1d23cd34b963555ac65abca21b081c


powerjg:
Yeah. Exactly. You can also use M5_VAR_USED if you really need to create a variable for it.

PR #18, opened 2018-08-22
flexcpu: Commit dependencies and strict serialization

animalvgamer:
Change-Id: I752172023a0440e21e9e5963140fc9d89670a329

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_cpu.hh
+++ b/src/cpu/flexcpu/simple_dataflow_cpu.hh
@@ -186,6 +186,14 @@ class SimpleDataflowCPU : public BaseCPU
     // END Internal constants
 
 
+    // BEGIN Internal parameters
+
+    // TODO temporary mark all executions as taking one cycle's length at least
+    Tick executionTime = clockPeriod();
+
+    // END Internal parameters
+
+
     // BEGIN Internal state variables
 
     DataPort _dataPort;
@@ -210,9 +218,6 @@ class SimpleDataflowCPU : public BaseCPU
     //       based on what has been learned since the writing of this though.
     std::unordered_map<PacketPtr, FetchCallback> outstandingFetches;
     std::unordered_map<PacketPtr, MemAccessReq> outstandingMemReqs;
-
-    // TODO temporary mark all executions as taking one cycle's length at least
-    Tick executionTime = clockPeriod();
 
     // END Internal state variables
 
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -44,10 +44,12 @@ using namespace std;
 // Fullsystem mode constructor
 SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
                     System* system_, BaseTLB* itb_, BaseTLB* dtb_,
-                    TheISA::ISA* isa_, bool use_kernel_stats_):
+                    TheISA::ISA* isa_, bool use_kernel_stats_,
+                    bool strict_ser):
+    memIface(*this),
     _cpuPtr(cpu_),
     _name(cpu_->name() + ".thread" + to_string(tid_)),
-    memIface(*this),
+    strictSerialize(strict_ser),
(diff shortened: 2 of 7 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -642,23 +643,55 @@ SDCPUThread::populateDependencies(shared_ptr<InflightInst> inst_ptr)
...
     for (auto& prior : inflightInsts) {
-        if (prior == inst_ptr || prior->isComplete() || prior->isSquashed()) {
+        if (prior == inst_ptr || prior->isCommitted() || prior->isSquashed()) {
             continue;
         }
 
-        const bool they_ser_after = prior->staticInst()->isSerializeAfter();
-        if (we_ser_before || they_ser_after) inst_ptr->addDependency(prior);
+        if (prior->staticInst()->isSerializeAfter())
+            inst_ptr->addCommitDependency(prior);
```
powerjg: Curly braces :)

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -642,23 +643,55 @@ SDCPUThread::populateDependencies(shared_ptr<InflightInst> inst_ptr)
...
+     * Note: An assumption is being made that inst_ptr is at the end of the
+     *       inflightInsts buffer.
+     */
 
     const StaticInstPtr static_inst = inst_ptr->staticInst();
 
     // BEGIN ISA explicit serialization
 
-    const bool we_ser_before = static_inst->isSerializeBefore();
+#define SDCPU_STRICT_SERIALIZATION
```
powerjg: I don't like using ifdefs if we can avoid it. Should this be a SimObject parameter, maybe? That could make sense while we're testing things, at least.

animalvgamer: Oh that's a good idea, didn't think of making it configurable that way...

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: But it's already indented so nicely

src/cpu/flexcpu/simple_dataflow_thread.cc
powerjg: Just to follow up on this. You'll want to add something like the following line to the SimObject description file (SimpleDataflowCPU.py).

`strict_serialization = Param.Bool(True, "Controls behavior of serializing instructions")`

Then, in the constructor of the C++ class `SimpleDataflowCPU`, you can pass the parameter to the thread. You can access the parameter with `params->strict_serialization`.

animalvgamer: I may or may not have just woken up :b, so this will be something I'll be in to work on later, thanks

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -49,6 +49,7 @@ class SDCPUThread;
  */
 class SimpleDataflowCPU : public BaseCPU
 {
+    friend class SDCPUThread;
```
powerjg: :/ Rather than doing this, why not pass the parameter through to the thread?

This is creating that coupling between the CPU and thread that you didn't want :).

animalvgamer: I'll make this change tomorrow, I'm working on the miscreg thing right now

animalvgamer: I guess one thing I'm scared of is just the constructor exploding in size if this is done a lot...

powerjg: It shouldn't... If we end up with many more things that need to be passed through we can revisit this.

animalvgamer: I think I'll just get rid of the friend class, but either expose or add accessors to the stored parameter

animalvgamer: but on the other hand, serialization is really the thread's job, not the CPU's :l

PR #19, opened 2018-08-23
flexcpu: Added non-operand miscreg access support

animalvgamer:
Change-Id: Iaa086224356c454034fdf0b5ed221fca24069c2a

### DIFF
--- a/src/cpu/flexcpu/inflight_inst.cc
+++ b/src/cpu/flexcpu/inflight_inst.cc
@@ -41,11 +41,13 @@ using CCReg = TheISA::CCReg;
 using VecElem = TheISA::VecElem;
 
 InflightInst::InflightInst(ThreadContext* backing_context,
+                           TheISA::ISA* backing_isa,
                            MemIface* backing_mem_iface,
                            InstSeqNum seq_num,
                            const PCState& pc_,
                            StaticInstPtr inst_ref):
     backingContext(backing_context),
+    backingISA(backing_isa),
     backingMemoryInterface(backing_mem_iface),
     _status(Waiting),
     _seqNum(seq_num),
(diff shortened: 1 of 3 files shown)


src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -276,7 +323,10 @@ InflightInst::readVecRegOperand(const StaticInst* si, int op_idx) const
 TheISA::VecRegContainer&
 InflightInst::getWritableVecRegOperand(const StaticInst* si, int op_idx)
 {
-    // TODO this one may produce bad results, need to rethink this.
+    // TODO this one may produce bad results, need to rethink this, since if
```
powerjg: You could add an assert here and make sure that things match your expectations :)

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -470,16 +520,21 @@ InflightInst::setMiscRegOperand(const StaticInst* si, int dst_idx,
...
 
 void
 InflightInst::setMiscReg(int misc_reg, const MiscReg& val)
 {
-    panic("setMiscReg() not implemented!");
-    // TODO
+    // O3 hides multiple writes to the same misc reg during execution, but due
+    // to potential side effects of each access, I don't think we should be
+    // hiding them. Instead, we will replay each write sequentially during
+    // commit time.
```
powerjg: Hmm... I'm not sure this will make any difference. Since you're functionally executing all of the writes on the same tick and there's no timing done for the side effects I think it'll be the same either way. This seems safer and easier to implement, though.

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -30,6 +30,8 @@
 
 #include "cpu/flexcpu/inflight_inst.hh"
 
+#include <algorithm>
```
powerjg: I could be missing this, but I don't see where it's used.

animalvgamer: Oh right, I added it to use find, but it turned out I didn't use it I think

src/cpu/flexcpu/inflight_inst.cc
animalvgamer: Oh no, that's just something I'm fairly certain I have to change eventually, I was just fleshing out the comment

src/cpu/flexcpu/inflight_inst.cc
animalvgamer: I thought all of those things too

PR #20, opened 2018-08-24
flexcpu: Add some stats and callbacks

powerjg:
Change-Id: I8e0b48d1938a96cce83386287b970e46846aa572

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.hh
+++ b/src/cpu/flexcpu/simple_dataflow_thread.hh
@@ -399,6 +399,15 @@ class SDCPUThread : public ThreadContext
 
     // END Internal functions
 
+    // BEGIN statistics
+
+    Counter numInsts = 0;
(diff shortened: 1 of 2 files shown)


src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -492,8 +501,14 @@ class SDCPUThread : public ThreadContext
...
 
+    /**
+     * Register thread-specific stats.
+     * The BaseCPU calls this function in its regStats with the thread-
+     * specific name
+     *
+     * @param name from the CPU for this thread
+     */
     void regStats(const std::string &name) override
-    { _committedState->regStats(name); }
```
animalvgamer: Is there a missing semicolon here?

powerjg:
That's why you should always run tests before pushing :)

PR #22, opened 2018-08-24
cpu-flexcpu: Added tracing for debug-flag Exec

animalvgamer:
Change-Id: Ic9d8f2e8ccb0bd2d5f493181f3d77d0059a0687e

### DIFF
--- a/src/cpu/flexcpu/inflight_inst.cc
+++ b/src/cpu/flexcpu/inflight_inst.cc
@@ -56,6 +56,11 @@ InflightInst::InflightInst(ThreadContext* backing_context,
     staticInst(inst_ref);
 }
 
+InflightInst::~InflightInst()
+{
+    if (_traceData) delete _traceData;
+}
+
 void
 InflightInst::addCommitCallback(function<void()> callback)
 {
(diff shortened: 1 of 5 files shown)


src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -601,6 +606,8 @@ InflightInst::initiateMemRead(Addr addr, unsigned int size,
                               Request::Flags flags)
 {
     if (backingMemoryInterface) {
+        if (_traceData)
```
powerjg: One line or curly braces?

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -56,6 +56,11 @@ InflightInst::InflightInst(ThreadContext* backing_context,
     staticInst(inst_ref);
 }
 
+InflightInst::~InflightInst()
+{
+    if (_traceData) delete _traceData;
```
powerjg: Why do you have this if you also have deleteTraceData?

animalvgamer: For things like faults that discard the InflightInst, I want to make sure it's cleared. deleteTraceData is so I can clear it earlier if we know we won't need it anymore I guess

powerjg: I see. I would pick one or the other. Can we just have the one in the destructor?

src/cpu/flexcpu/inflight_inst.cc
animalvgamer: done

src/cpu/flexcpu/inflight_inst.cc
animalvgamer: done

PR #23, opened 2018-09-03
cpu-flexcpu: added an internal thread fetch-buffer

animalvgamer:
Change-Id: I83f07fd293c42dd211cbffc72319b5ba68319fc8

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -45,31 +45,42 @@ using namespace std;
 SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
                     System* system_, BaseTLB* itb_, BaseTLB* dtb_,
                     TheISA::ISA* isa_, bool use_kernel_stats_,
-                    bool strict_ser):
+                    unsigned fetch_buf_size, bool strict_ser):
     memIface(*this),
     _cpuPtr(cpu_),
     _name(cpu_->name() + ".thread" + to_string(tid_)),
     strictSerialize(strict_ser),
     _committedState(new SimpleThread(cpu_, tid_, system_, itb_, dtb_, isa_,
                                      use_kernel_stats_)),
-    isa(isa_)
+    isa(isa_),
+    fetchBuf(vector<uint8_t>(fetch_buf_size)),
+    fetchBufMask(~(static_cast<Addr>(fetch_buf_size) - 1))
 {
-    //
+    panic_if(fetch_buf_size % sizeof(TheISA::MachInst) != 0,
+             "Fetch buffer size should be multiple of instruction size!");
+    panic_if(fetch_buf_size > _cpuPtr->cacheLineSize(),
+             "Can't send fetch requests larger than cache lines!");
 }
 
 // Non-fullsystem constructor
 SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
                     System* system_, Process* process_, BaseTLB* itb_,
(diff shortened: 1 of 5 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -45,31 +45,42 @@ using namespace std;
...
                                      use_kernel_stats_)),
-    isa(isa_)
+    isa(isa_),
+    fetchBuf(vector<uint8_t>(fetch_buf_size)),
+    fetchBufMask(~(static_cast<Addr>(fetch_buf_size) - 1))
 {
-    //
+    panic_if(fetch_buf_size % sizeof(TheISA::MachInst) != 0,
+             "Fetch buffer size should be multiple of instruction size!");
+    panic_if(fetch_buf_size > _cpuPtr->cacheLineSize(),
```
powerjg: This is fine for now, but we might want the buffer to be larger than a cache line. If it is larger, then we just send multiple cache-line size requests as the buffer is drained.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -158,7 +169,96 @@ SDCPUThread::advanceInst()
     inflightInsts.push_back(dynamic_inst_ptr);
 
     fetchOffset = 0;
-    onBeginFetch(weak_ptr<InflightInst>(dynamic_inst_ptr));
+    attemptFetch(weak_ptr<InflightInst>(dynamic_inst_ptr));
+}
+
+void
+SDCPUThread::attemptFetch(weak_ptr<InflightInst> inst)
```
powerjg: I don't like this name...

Fetching is when you get data from memory. Decoding is when you decode... This is moving the instruction that has been fetched into the decoder. I'm not sure what the name *should* be. 

Why is this a separate step? Why doesn't the decoder just get data directly from the fetch buffer?

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -194,12 +205,30 @@ class SDCPUThread : public ThreadContext
...
 
+    /**
+     * Entry point to fetch functionality. May be called more than once for a
+     * particular instruction, due to instructions spanning across fetch
+     * boundaries.
+     *
+     * @param inst A reference to the in-flight instruction object for which to
+     *  request fetch data.
+     */
+    void attemptFetch(std::weak_ptr<InflightInst> inst);
```
powerjg: Yeah, I think this should be combined with the decode logic.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: Hmm, my thought when deciding on the limits was to use this more like the DRAM buffer where we just had the capability to hold the result of one access at a time until we needed something else. I was afraid of going beyond that, because there would come a point where it might make more sense to model any nuances with the instruction cache by attaching a different icache to the instruction port...

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: This was just the onBeginFetch function renamed. When I made this feature change, I tried to do it in a way that minimized the change from the flow logic before. Since the buffer is functionally like an internal icache, it now includes logic to check that "cache" before sending a response to the external icache.

src/cpu/flexcpu/simple_dataflow_thread.hh
animalvgamer: The other thing is that a decode could occur across the boundaries of two fetches, even with the buffer, so I think trying to combine it might require a more significant logic change.

PR #27, opened 2018-09-07
Jason/flexcpu/clocked

powerjg:
This is another way to do it :). I think it's a little cleaner.

Feel free to clean up by putting things in alphabetical order.

Also, I didn't run many tests (just the regressions), and I'd like to dig into make sure the code makes sense a bit more, too. I'm not sure why we would ever not be on a clock edge...

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_cpu.cc
+++ b/src/cpu/flexcpu/simple_dataflow_cpu.cc
@@ -47,6 +47,7 @@ SimpleDataflowCPU::SimpleDataflowCPU(SimpleDataflowCPUParams* params):
     clockedInstFetch(params->clocked_inst_fetch),
     clockedItbTranslation(params->clocked_itb_translation),
     clockedMemoryRequest(params->clocked_memory_request),
+    executionLatency(params->execution_latency),
     _dataPort(name() + "._dataPort", this),
     _instPort(name() + "._instPort", this)
 {
@@ -122,7 +123,8 @@ SimpleDataflowCPU::attemptAllExecutions()
             completeExecution(req);
         }, name() + ".executeEvent", true);
 
-        schedule(event, executionTime);
+        schedule(event, clockedExecution ? clockEdge(executionLatency) :
+                 curTick() + cyclesToTicks(executionLatency));
 
         executionReqs.pop_front();
     }
--- a/src/cpu/flexcpu/simple_dataflow_cpu.cc
+++ b/src/cpu/flexcpu/simple_dataflow_cpu.cc
@@ -605,6 +605,20 @@ SimpleDataflowCPU::startup()
 }
 
 void
+SimpleDataflowCPU::suspendContext(ThreadID tid)
+{
+    // Need to de-schedule any instructions in the pipeline?
+    BaseCPU::suspendContext(tid);
+}
+
+void
+SimpleDataflowCPU::haltContext(ThreadID tid)
+{
+    // Need to de-schedule any instructions in the pipeline?
+    BaseCPU::haltContext(tid);
+}
+
+void
 SimpleDataflowCPU::switchOut()
 {
     // TODO flush speculative state.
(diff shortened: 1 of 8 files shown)


powerjg:
@ayaz91 Feel free to start reviewing the code! More eyes are always better!

animalvgamer:
Is it okay to make changes to the commit on your branch? It'd be the only way to make changes and keep the same pull request

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -597,7 +672,11 @@ void
 SimpleDataflowCPU::DataPort::recvReqRetry()
 {
     receivedPushback = false;
-    cpu->attemptAllMemReqs();
+
+    if (!cpu->attemptAllMemReqsEvent.scheduled()) {
+        cpu->schedule(&cpu->attemptAllMemReqsEvent,
+            cpu->clockedMemoryRequest ? cpu->clockEdge(Cycles(0)) : curTick());
+    }
```
animalvgamer: I would add comments to each of these points, mentioning the exact function that these events are going to call, like attemptAllMemReqs(), since that will make control flow easier to follow.

animalvgamer:
And to be wholly fair, to answer the not being on a clock edge thing, I really think these changes only make a big difference in the case where any task takes some amount of time to complete that isn't specifically an integer number of cycles. I became less convinced that this change was even sensible over time as I wrote it, since I think the goal of this change was just to facilitate the addition of:
  1) counters so IPC could be more reasonably tracked
  2) constraints so that access to things could be limited on a per-cycle basis

Both these goals might be better achieved by just implementing the resource limits and the reset event, and counting the number of resets as cycles, such that any event processed during a cycle belong to the cycle in between these reset events.

I don't think IPC being tracked beyond an (instructions/unit time) measure is valuable right now without said constraints however.

As an aside, I honestly think that this may have been a bad route to take, and didn't really add anything new or interesting to the project. I think that pursuing the speed-increasing features like Branch Prediction and relaxing the memory model are going to be the most valuable investments of time right now, since we have to be able to run faster than O3 first to be able to write custom constraints to help us match O3 by configuration alone.

powerjg:
I mostly agree :). Also, feel free to edit my branch.

Jason

animalvgamer:
Also are these commits going to be kept separate or squashed before merging?

animalvgamer:
Is it possible to separate some of these changes into different branches, so we can have separate reviews? There are a number of changes I would accept already, but maybe not as a whole...

powerjg:
What's in here that we can't accept ;).

Feel free to cherry pick whatever you want to on flexcpu. I'll deal with
the rebase headache. All of these patches have been helping me work on the
issue I found after adding clocking.

animalvgamer:
Just the alphabetical ordering right now. My last comment was just so that this changeset doesn't get too big before the merge.

powerjg:
Ok. Fixed ordering and changes latency to cycles. I think it should be cycles.

powerjg:
nvm. This rebase is hell.

powerjg:
OK! I think I'm happy with this PR now.

powerjg:
Let me know if there's anything that needs to be updated here. I'd like to merge today.

animalvgamer:
Is it possible to set latency to ticks instead of cycles (and default to one cycle's worth of ticks)? I think that would best facilitate keeping clocking as a parameter.

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -73,10 +79,62 @@ SimpleDataflowCPU::activateContext(ThreadID tid)
...
+    assert(!executionReqs.empty());
+
+    while (!executionReqs.empty()) {
+        auto req = executionReqs.front();
+
+        Event* event = new EventFunctionWrapper([this, req]() {
+            completeExecution(req);
+        }, name() + ".executeEvent", true);
+
+        schedule(event, clockEdge(executionLatency));
```
animalvgamer: This should respect the parameter clocked_execution. If it's not clocked_execution, it should probably be curTick() + clockPeriod() * executionLatency or something like that.

animalvgamer: or curTick() + cyclesToTicks(executionLatency)

animalvgamer: they're actually the same, just one is in-lined I guess

PR #29, opened 2018-09-11
Flexcpu

powerjg:
General PR for making comments on the entire project. Let's see how this works...


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,1050 @@
...
+
+    // Retrieve the value of the speculative PC as a raw Addr value, aligned
+    // to MachInst size, but offset by the amount of data we've fetched thus
+    // far, since multiple fetches may be needed to feed one decode.
+    const Addr pc_addr = pc_value.instAddr();
+    const Addr req_vaddr = (pc_addr & BaseCPU::PCMask) + fetchOffset;
+
+    // If HIT in fetch buffer, we can skip sending a new request, and decode
+    // from the buffer immediately.
+    if (fetchBufBase
```
powerjg: I think the fetch buffer should be in the CPU, not in the thread. This way you can check for bandwidth/latency on all fetches.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -0,0 +1,1050 @@
...
+    const unsigned req_size = fetchBuf.size();
+
+    fetch_req->setVirt(0, new_buf_base, req_size,
+                       Request::INST_FETCH, _cpuPtr->instMasterId(), pc_addr);
+
+    auto callback = [this, inst](Fault f, const RequestPtr& r) {
+        onPCTranslated(inst, f, r);
+    };
+
+    if (!_cpuPtr->requestInstAddrTranslation(fetch_req, getThreadContext(),
```
powerjg: It's a little weird that you don't translate things that hit in the fetch buffer. It could be a problem for two reasons:
1) In some sense you're "speculating" that it's a TLB hit. It might not be. If you are going to skip the translation, then you need a mechanism to *guarantee* that it's a TLB hit.
2) From a software point of view, you're sometimes bypassing the TLB and sometimes not. That's also a little weird. It makes the software logic more complicate (e.g., if you try to explain it with a diagram).

animalvgamer:
Are the first three commits supposed to be here? Have they not been merged into master yet?

powerjg:
Yeah... I guess merging master will be a pain because of that. We could rebase this on master, but that would be an even bigger pain....

powerjg:
I'm going to close this PR. We can make a new PR with the renamed/updated branches.

PR #30, opened 2018-09-13
Jason/flexcpu/resources

powerjg:
Most of the bandwidth stuff is there now. The only thing that isn't is tracking the number of "executes" at each resource and resetting that number every cycle. The current implementation essentially works the same was before the patch, but now it's set up to limit bandwidth. 

I'm posting now to get feedback. Also, I'm going to have to take time off from coding for the next couple of days, so if someone else wants to finish this PR off, that would be great!

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_cpu.cc
+++ b/src/cpu/flexcpu/simple_dataflow_cpu.cc
@@ -46,7 +46,7 @@ SimpleDataflowCPU::SimpleDataflowCPU(SimpleDataflowCPUParams* params):
     dataAddrTranslationUnit(this, params->clocked_dtb_translation, Cycles(0),
                             0, name() + ".dtbUnit"),
     executionUnit(this, params->clocked_execution, params->execution_latency,
-                  0, name() + ".executionUnit"),
+                  params->execution_bandwidth, name() + ".executionUnit"),
     instFetchUnit(this, params->clocked_inst_fetch, 0, name() + ".iFetchUnit"),
     instAddrTranslationUnit(this, params->clocked_itb_translation, Cycles(0),
                             0, name() + ".itbUnit"),
@@ -727,7 +727,8 @@ SimpleDataflowCPU::Resource::Resource(SimpleDataflowCPU *cpu, bool clocked,
                     name() + ".attemptAll", false,
                     run_last ? Event::CPU_Tick_Pri : Event::Default_Pri))
 {
-
+    fatal_if(latency == 0 && bandwidth > 0 && !clocked,
+             "Cannot have 0 latency and finite bandwidth unless clocked.");
 }
 
 void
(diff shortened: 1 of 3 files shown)


src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -692,12 +642,29 @@ SimpleDataflowCPU::DataPort::recvTimingResp(PacketPtr pkt)
 void
 SimpleDataflowCPU::DataPort::recvReqRetry()
 {
-    receivedPushback = false;
+    assert(blockedPkt);
+    DPRINTFS(SDCPUCoreEvent, cpu, "data retrying %s\n", blockedPkt->print());
```
animalvgamer: Was there something wrong with letting the name of the port be used to DPRINTF before?

src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -727,12 +694,145 @@ SimpleDataflowCPU::InstPort::recvTimingResp(PacketPtr pkt)
...
+{
+    fatal_if(latency != Cycles(0), "Memory resources must have 0 latency!");
+}
+
+SimpleDataflowCPU::InstFetchResource::InstFetchResource(SimpleDataflowCPU *cpu,
+    bool clocked, Cycles latency, int bandwidth, std::string _name,
+    bool run_last) :
+    Resource(cpu, clocked, latency, bandwidth, _name, run_last)
+{
+    fatal_if(latency != Cycles(0), "Memory resources must have 0 latency!");
```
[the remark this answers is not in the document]
animalvgamer: Also I think it's probably not that weird for memory access to have latency additional to what the port system causes.

src/cpu/flexcpu/simple_dataflow_cpu.hh
```diff
@@ -192,33 +178,134 @@ class SimpleDataflowCPU : public BaseCPU
...
+     */
+    class Resource
+    {
+      protected:
+        SimpleDataflowCPU* cpu;
+
+      private:
+        const bool clocked;
+        const Cycles latency;
+        const int bandwidth;
```
animalvgamer: I don't think this gives us enough information to model what we want to model. I'd like to try to draw this out in person.

animalvgamer: Like I think we mentioned it a little when discussing how to limit latency and bandwidth, but I think we also need to think about pipelinability?

animalvgamer: I think this part shouldn't stop us from merging the patch, since not merging now will make this very large changeset harder to merge, but it will probably look pretty different by the time we fully implement constraints in their final form.

powerjg: I agree with this :).

PR #34, opened 2018-09-19
Jason/flexcpu/issue resource

powerjg:
PR for now. This still needs to move how advanceInst is called... or something :)

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -132,75 +132,152 @@ SDCPUThread::advanceInst(TheISA::PCState next_pc)
 
     const InstSeqNum seq_num = lastCommittedInstNum + inflightInsts.size() + 1;
 
-    DPRINTF(SDCPUThreadEvent, "advanceInst(seq %d, pc %#x, upc %#x)\n",
+    DPRINTF(SDCPUThreadEvent, "advanceInst(seq %d, pc %#x, upc %d)\n",
                               seq_num,
                               next_pc.pc(),
                               next_pc.upc());
-    DPRINTF(SDCPUThreadEvent, "Buffer size increased to %d\n",
-                              inflightInsts.size());
+
+    StaticInstPtr static_inst = nullptr;
 
     if (isRomMicroPC(next_pc.microPC())) {
-        const StaticInstPtr static_inst =
+        assert(!curMacroOp);
+        static_inst =
             _cpuPtr->microcodeRom.fetchMicroop(next_pc.microPC(), curMacroOp);
 
         DPRINTF(SDCPUInstEvent,
-                "Decoded ROM microop (seq %d) - %#x : %s\n",
+                "Decoded ROM microop (seq %d) - %d : %s\n",
                 seq_num,
                 next_pc.microPC(),
                 static_inst->disassemble(next_pc.microPC()).c_str());
-
-        shared_ptr<InflightInst> dynamic_inst_ptr =
-            make_shared<InflightInst>(getThreadContext(), isa, &memIface,
-                                      seq_num, next_pc, static_inst);
-
-        inflightInsts.push_back(dynamic_inst_ptr);
-
-        issueInstruction(dynamic_inst_ptr);
-
-        return;
-    }
-
-    if (curMacroOp) {
-        const StaticInstPtr static_inst =
-            curMacroOp->fetchMicroop(next_pc.microPC());
+    } else if (curMacroOp) {
+        static_inst = curMacroOp->fetchMicroop(next_pc.microPC());
 
         DPRINTF(SDCPUInstEvent,
-                "Decoded microop (seq %d) - %#x : %s\n",
+                "Decoded microop (seq %d) - %d : %s\n",
                 seq_num,
                 next_pc.microPC(),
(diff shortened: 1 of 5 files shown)


powerjg:
Ugh... my rebasing went horribly wrong. However, I think you can see my changes here. I'll fix the rebase later.

powerjg:
Ok. This is looking better now. I'd like feedback on the re-org and other changes. I'm still working on actually constraining the issue. Not having perfect branch prediction is making it hard.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -132,76 +132,153 @@ SDCPUThread::advanceInst(TheISA::PCState next_pc)
...
                               next_pc.pc(),
                               next_pc.upc());
-    DPRINTF(SDCPUThreadEvent, "Buffer size increased to %d\n",
-                              inflightInsts.size());
+
+    StaticInstPtr static_inst = nullptr;
 
     if (isRomMicroPC(next_pc.microPC())) {
-        const StaticInstPtr static_inst =
+        assert(!curMacroOp);
```
animalvgamer: This assert is not valid, since we actually use the macroOp to extract things from the microROM.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -270,8 +328,8 @@ SDCPUThread::bufferInstructionData(Addr vaddr, uint8_t* data)
     DPRINTF(SDCPUThreadEvent, "Updating fetch buffer with data from %#x\n",
                               vaddr);
 
+    assert(vaddr);
     fetchBufBase = vaddr;
-    if (!vaddr) return;
```
animalvgamer: It was done this way before, to provide a function call which would clear the buffer.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -754,9 +752,16 @@ SDCPUThread::onExecutionCompleted(weak_ptr<InflightInst> inst, Fault fault)
                     squashUpTo(inst_ptr, true);
 
                     // Notify branch predictor of incorrect prediction
-
-                    _cpuPtr->getBranchPredictor()->squash(inst_ptr->seqNum(),
```
animalvgamer: I don't think this was the right squash function to call -- probably should just call the other squash function right now.

animalvgamer: Note that if you add a commit callback, the callback will never be called, since the faulting instruction is not committed.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -856,17 +805,18 @@ SDCPUThread::onPCTranslated(weak_ptr<InflightInst> inst, Fault fault,
     DPRINTF(SDCPUThreadEvent, "Received PC translation (va: %#x, pa: %#x)\n",
             vaddr, req->getPaddr());
 
-    auto callback = [this, inst, vaddr](uint8_t* data) {
-        bufferInstructionData(vaddr, data);
-        attemptFetch(inst);
+    auto callback = [this, next_pc, vaddr](uint8_t* data) {
+        if (fetchPC == next_pc) {
```
animalvgamer: Didn't we run into issues with this if you had two outstanding requests to the same PC?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -1053,12 +1003,14 @@ SDCPUThread::predictCtrlInst(shared_ptr<InflightInst> inst_ptr)
 {
     DPRINTF(SDCPUBranchPred, "Requesting branch predictor for control\n");
 
-    const InstSeqNum seqnum = inst_ptr->seqNum();
-    inst_ptr->addSquashCallback([this, seqnum] {
-        _cpuPtr->getBranchPredictor()->squash(seqnum, threadId());
+    const weak_ptr<InflightInst> weak_inst = inst_ptr;
+    inst_ptr->addSquashCallback([this, weak_inst] {
```
animalvgamer: This is redundant, because this callback belongs to inst_ptr. weak_inst will never fail to lock, otherwise we would be unable to call the squash callback. We didn't capture inst_ptr either because we didn't need a reference to it to call squash, as long as we had the sequence number...

animalvgamer: It's a more lightweight capture to just capture an unmanaged reference to a SDCPUThread and the seqnum than a managed weak_ptr.

animalvgamer: and the callback will already naturally be cleaned up or not called, since it's purely managed by the inst_ptr's callback list.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: I think a better way to manage this is to keep a flag/management object for outstanding requests to the CPU, and to kill the object and the corresponding request during any squash. Then restart advancing. This was one of the jobs the empty shared_ptr<InflightInst> served when it was in the buffer, which is still an important job to do, even if we don't like having InflightInsts early.

powerjg:
I'm going to close this PR since the changes are in the other PR (#38) anyway.

PR #35, opened 2018-09-19
[fixup] use the right backing memory for split request packets

animalvgamer:


animalvgamer:
The pointer parameter uint8_t* data will no longer be valid upon return of the function, meaning that if we assign it to the pointers as a static pointer, the data going to memory could be overwritten by the time the packet is handled.

powerjg:
Makes sense! Merge whenever :).

PR #36, opened 2018-09-24
Conservative out-of-order loads, assuming dependencies exist until all previous stores have calculated effective address.

animalvgamer:


src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -109,6 +112,78 @@ InflightInst::addDependency(shared_ptr<InflightInst> parent)
...
+    if (parent->isSquashed() || parent->isComplete()) return;
+
+    ++remainingMemDependencies;
+
+    weak_ptr<InflightInst> weak_this = shared_from_this();
+    parent->addCompletionCallback([weak_this]() {
+        shared_ptr<InflightInst> inst_ptr = weak_this.lock();
+        if (!inst_ptr) return;
+
+        --inst_ptr->remainingMemDependencies;
```
powerjg: General comment: This code (subtract dependencies and test for notify) is repeated multiple times. I think it should be a member function of InflightInst.

In fact, looking at this code... maybe we shouldn't be using so many lambdas here. It might make more sense to "hard code" the behavior for commit/mem/complete and use lists of instructions instead of lists of functions. I'm mostly saying this so we have a record of it for later.

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -55,15 +55,20 @@ class InflightInst : public ExecContext,
     using VecRegContainer = TheISA::VecRegContainer;
     using VecElem = TheISA::VecElem;
 
+    // Mutually exclusive states through which InflightInsts will transition.
```
powerjg: Lol. These need to be all the same part of speech. I don't have concrete suggestions, though.

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -124,10 +129,16 @@ class InflightInst : public ExecContext,
     // Who depends on me?
     std::vector<std::function<void()>> completionCallbacks;
 
-    // How many dependencies have to be resolved before I can run?
+    std::vector<std::function<void()>> effAddrCalculatedCallbacks;
```
powerjg: This is getting a little complicated with 4 different kinds of callbacks and two counters. It might be a good time to document why all of this is necessary. (This is something that could also go in the paper under the design section, so it's not wasted time to write it.)

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -204,17 +261,28 @@ class InflightInst : public ExecContext,
     }
 
     inline bool isCommitted() const
-    { return status() == Committed; }
+    { return status() >= Committed; }
```
powerjg: Need some comment about how the order now matters. Need comments both at the enum definition and here to say committed implies effAddred, etc.

src/cpu/flexcpu/inflight_inst.hh
```diff
@@ -204,17 +261,28 @@ class InflightInst : public ExecContext,
...
-    { return status() == Complete; }
+    { return status() >= Complete; }
+
+    inline bool isEffAddred() const
+    { return status() >= EffAddred; }
 
     inline bool isFaulted() const
     { return fault() != NoFault; }
 
+    inline bool isMemReady()
```
powerjg: What does this mean? If `isMemReady()` does that mean that you can send it to memory? Or...?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -969,43 +980,136 @@ SDCPUThread::populateDependencies(shared_ptr<InflightInst> inst_ptr)
...
+    // BEGIN Conservative Memory dependence ordering
+
+    if (static_inst->isLoad()) {
+        // TODO check all previous memory referencing instructions.
+        //      if effaddr resolved, add self check to see if memory needs
+        //         dependence.
+        //      else add effaddr dependence.
+        weak_ptr<InflightInst> weak_inst = inst_ptr;
+
+        for (shared_ptr<InflightInst>& other : inflightInsts) {
```
powerjg: Remind me what order this is (in a comment). Are we starting with the most youngest or oldest instruction?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -1081,11 +1187,77 @@ SDCPUThread::recordCycleStats()
...
+    if (!inst_ptr || inst_ptr->isSquashed()) {
+        // No need to do anything for an instruction that has been squashed.
+        return;
+    }
+
+    sendToMemory(inst_ptr, req, write, data, sreq);
+}
+
+void
+SDCPUThread::sendToMemory(shared_ptr<InflightInst> inst_ptr,
```
powerjg: I'm a little surprised this compiles. I thought that `shared_ptr` could be implicitly cast to `weak_ptr`. How does the compiler know which one to call? Even if the compiler *can* infer the right function, It's confusing for people reading the code.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -1095,10 +1267,34 @@ SDCPUThread::squashUpTo(shared_ptr<InflightInst> inst_ptr, bool rebuild_map)
 
     DPRINTF(SDCPUThreadEvent, "%d instructions squashed.\n", count);
 
-    if (rebuild_map) {
+    if (rebuild_lasts) {
         for (shared_ptr<InflightInst>& it : inflightInsts) {
             populateUses(it);
         }
+
+        for (auto itr = inflightInsts.rbegin();
```
powerjg: What a pain. But I see why it has to be done.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -466,6 +475,51 @@ class SDCPUThread : public ThreadContext
...
+     * @param inst A reference to the instruction for which to request memory.
+     * @param req The request object which non-split requests will use.
+     * @param write Whether the request is a write. If not write, then read.
+     * @param data The data to write, if a write is requested.
+     * @param sreq A split request structure containing the divided and
+     *  translated requests.
+     */
+    void sendToMemory(std::weak_ptr<InflightInst> inst,
+                      const RequestPtr& req, bool write,
+                      std::shared_ptr<uint8_t> data,
```
powerjg: Could this have been a `vector<uint8_t>`. Seems like that would be safer since it's a pointer to data, not a single `uint8_t`.

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -70,6 +70,8 @@ InflightInst::addCommitCallback(function<void()> callback)
 void
 InflightInst::addCommitDependency(shared_ptr<InflightInst> parent)
 {
+    if (parent->isSquashed() || parent->isCommitted()) return;
```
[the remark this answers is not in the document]
powerjg: Actually, I guess I see that you have moved this check into the dependency functions so you don't have to do it every time you call the function. Ignore the above comment.

src/cpu/flexcpu/inflight_inst.cc
animalvgamer: Well it has to do with the fact that it's different stages that determine one stage's dependencies. If you wanted to hard code the behavior, it'd still have to be something along the lines of "decrementAndCheckMemDependencies()" which would be called for every child of the parent, instead of through the single for loop in notifyX().

animalvgamer: The lambda allows us the capture a reference to "this" and store it in the parent.

src/cpu/flexcpu/inflight_inst.hh
animalvgamer: They're all adjectives?

src/cpu/flexcpu/inflight_inst.hh
animalvgamer: Yea I could add more comments, but fundamentally, the callbacks are for anything that needs to be called during a state transition, and dependencies are things that have to be satisfied before a state transition.

src/cpu/flexcpu/inflight_inst.hh
animalvgamer: Added

src/cpu/flexcpu/inflight_inst.hh
animalvgamer: Yes. Well, specifically that there are no longer any dependencies tracked by this InflightInst that would prevent it from going to memory. There are some minor nuances as to when it's sent to memory depending on how the dependencies are set up, but otherwise that is the core semantic.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: front is oldest, so we are going from oldest to youngest in this case. Order doesn't matter in this case though, because we are checking all older stores.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: Well it's implicitly castable but it's still a different type... the compiler will always try to match exactly first, then try any casts, so if there were two things that a shared_ptr could be casted to, then it would probably complain.

This was something I thought to do after you showed me the perf details because there are cases where I call sendToMemory with a weak_ptr and with a shared_ptr. This overload allows us to make both calls safely, while minimizing the number of extra weak_ptrs being constructed, and the number of shared_ptr being created from .lock().

src/cpu/flexcpu/simple_dataflow_thread.hh
animalvgamer: :T well you'd be copying that data twice in the case of a split access, and multiple times since we're passing it by value through lambdas in order to allow it to survive to calls made in different ticks...

PR #38, opened 2018-09-25
Jason/flexcpu/performance improvement

powerjg:


src/cpu/flexcpu/simple_dataflow_cpu.cc
```diff
@@ -204,17 +206,19 @@ SimpleDataflowCPU::requestBranchPredictor(
 
 void
 SimpleDataflowCPU::requestDataAddrTranslation(const RequestPtr& req,
-    ThreadContext* tc, bool write,
+    ThreadContext* tc, bool write, std::weak_ptr<ExecContext> context,
```
animalvgamer: This is where I think a dedicated management block might be useful, since we're creating a typed weak_ptr without actually using the type at all.

A better alternative in this specific case I think is std::weak_ptr<void>, because the communicates that we're using the weak_ptr aspect of it to control the request, and we don't care at all what the pointer points to because we don't use it.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -122,6 +122,35 @@ SDCPUThread::activate()
 
 void
 SDCPUThread::advanceInst(TheISA::PCState next_pc)
+{
+    DPRINTF(SDCPUThreadEvent, "Request issue for %s. outstanding %p\n",
+            next_pc, squashIssueFlag);
+
+    squashIssueFlag = new bool(false);
```
animalvgamer: Could just make this a `shared_ptr<bool>`, since it's basically a pointer that's shared between SDCPUThread and the two lambdas using it here. When each reference dies it will clean up without having to manually do deletes here.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -132,76 +161,153 @@ SDCPUThread::advanceInst(TheISA::PCState next_pc)
 
     const InstSeqNum seq_num = lastCommittedInstNum + inflightInsts.size() + 1;
 
-    DPRINTF(SDCPUThreadEvent, "advanceInst(seq %d, pc %#x, upc %#x)\n",
+    DPRINTF(SDCPUThreadEvent, "advanceInst(seq %d, pc %#x, upc %d)\n",
```
animalvgamer: This should be changed to issue if that's where the DPRINTF is...

src/cpu/flexcpu/simple_dataflow_thread.cc
powerjg: Sure. That makes sense.

src/cpu/flexcpu/simple_dataflow_thread.cc
powerjg: Yeah. That was a quick change this morning/last night as I was running out of time to work on this.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -122,6 +122,35 @@ SDCPUThread::activate()
...
+
+    bool *squash_flag = squashIssueFlag;
+    _cpuPtr->requestIssue(
+        [this, next_pc, squash_flag] {
+            assert(!(*squash_flag));
+            if (squashIssueFlag == squash_flag) squashIssueFlag = nullptr;
+            delete squash_flag;
+            issueInst(next_pc);
+        },
+        [this, squash_flag] {
```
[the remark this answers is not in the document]
animalvgamer: Using a shared_ptr stored in the inflightInsts buffer gave us two major advantages -- ordering and unification.

1) Each request being tied to one has a place in the list of tasks the CPU has, based on where its corresponding shared_ptr in the buffer is.
2) By tying the requests to the shared_ptr in the buffer, you can centralize squashing logic such that you only have to make the check `if (!inst_ptr || inst_ptr->squashed() || inst_ptr->isAlreadyInALaterStageThanThisRequestIsFor())`. That way without having to know what's outstanding, you can squash all requests at once by squashing the instruction itself. A should_process boolean function could also make use of this logic, if the shared_ptr was placed back into the buffer, which is why I don't see why this implementation of advance/issue can't have the inst be created earlier for this exact management purpose.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -132,76 +161,153 @@ SDCPUThread::advanceInst(TheISA::PCState next_pc)
...
+    DPRINTF(SDCPUThreadEvent, "advanceInst(seq %d, pc %#x, upc %d)\n",
                               seq_num,
                               next_pc.pc(),
                               next_pc.upc());
-    DPRINTF(SDCPUThreadEvent, "Buffer size increased to %d\n",
-                              inflightInsts.size());
+
+    StaticInstPtr static_inst = nullptr;
 
     if (isRomMicroPC(next_pc.microPC())) {
```
animalvgamer: This function has gotten really really long with all the different things joined into it hmm... I think some of this could be moved into a different utility/helper function like decodeInst() for code organization if for nothing else.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -122,6 +122,35 @@ SDCPUThread::activate()
```
powerjg: Agreed.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: Basically a named step that generates a StaticInstPtr.

Basically we have 3 steps:
  - Check if we have data (either in a macroop or in the fetch buffer) and fetch if we don't <- kinda advanceInst before
  - Decode if we do (from the macroop or the fetch buffer) <- kinda onInstDataFetched before
  - Setup the InflightInst state like dependencies <- already a function

Then a call to the next step in the pipeline (execute if ready)

animalvgamer: It also kind of bothers me that advanceInst/issueInst aren't a one-to-one relationship with InflightInst anymore in this implementation. Basically every other event was always one-to-one, except for the data fetch loop before, since each function represented an explicit step taken towards committing a specific instruction (hence each function taking an instruction as an argument and operating on it).

Separating out the "named step that generates a StaticInstPtr" from advanceInst/issueInst I think would be a good way to restore that, if that's something we want to maintain.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -132,76 +161,153 @@ SDCPUThread::advanceInst(TheISA::PCState next_pc)
...
-        return;
-    }
-
     DPRINTF(SDCPUThreadEvent, "attemptFetch()\n");
 
+    // Save the PC we are trying to fetch from. It's possible that a branch
+    // misprediction will cause us to fetch from a different PC before this
+    // memory request returns. If, when this request is done, the fetch PC
+    // is different, then we need to just ignore this data.
+    fetchPC = next_pc;
```
animalvgamer: I wonder if that is a safe assumption to make -- that if a request response matches the PC we want, then it's always safe to accept it?

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -433,15 +433,15 @@ class SDCPUThread : public ThreadContext
...
+    void populateUses(std::shared_ptr<InflightInst>& inst_ptr);
 
     /**
      * This function makes a request for the branch predictor for the given
      * control instruction.
      *
      * @param inst_ptr The control instruction which needs a prediction.
      */
-    void predictCtrlInst(std::shared_ptr<InflightInst> inst_ptr);
+    void predictCtrlInst(std::shared_ptr<InflightInst>& inst_ptr);
```
animalvgamer: Pretty sure this can actually be `const std::shared_ptr<InflightInst>& inst_ptr`. Since it's the pointer that's const, not the InflightInst it references. If it were `std::shared_ptr<const InflightInst>& inst_ptr` that would be a different story.

PR #39, opened 2018-09-26
Merged loooad and the BP changes.

animalvgamer:


powerjg:
This looks OK and I think it's best to merge it as is. We need to add back the execution latency parameter that I forgot to add back, but that can be another patch.

The slow performance worries me some. Have you checked to see if the branch predictor is correctly being skipped on incorrect branches? It should easy enough to check the debug log for the first few hundred cycles. It would also be good to add the assert in freeBranchPredDepth to make sure that it's being handled correctly.

animalvgamer:
Okay, merging then.

As for the performance, observations I've made so far are that the buffer has a tendency to get extremely large with this version of flexcpu. It will commonly hit a size of 900+ on IntMM and FloatMM, and what happens is that small squashes of ~10 instructions occur, but I'm guessing it just takes such a long time to rebuild the lasts when these squashes occur because of the size of the buffer... I am currently exploring whether this is "correct" behavior, since it does match stdout output.

PR #47, opened 2018-10-04
cpu-flex: allowed independent out-of-order stores

animalvgamer:
Change-Id: I2ba5944869729ffddf20b9bc128a0342eda50a73

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -292,6 +292,14 @@ SDCPUThread::bufferInstructionData(Addr vaddr, uint8_t* data)
     memcpy(fetchBuf.data(), data, fetchBuf.size());
 }
 
+bool
+SDCPUThread::canCommit(shared_ptr<InflightInst> inst_ptr)
+{
+    return !inst_ptr->isCommitted() &&
+           (inst_ptr->isComplete()
+            || (inst_ptr->isMemorying() && inst_ptr->staticInst()->isStore()));
+}
+
 void
 SDCPUThread::commitAllAvailableInstructions()
 {
@@ -302,7 +310,7 @@ SDCPUThread::commitAllAvailableInstructions()
 
     bool buffer_shrunk = false;
     while (!inflightInsts.empty()
-        && ((inst_ptr = inflightInsts.front())->isComplete()
+        && (canCommit(inst_ptr = inflightInsts.front())
             || inst_ptr->isSquashed())) {
 
         if (inst_ptr->fault() != NoFault) {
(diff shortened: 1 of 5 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -1380,35 +1386,61 @@ SDCPUThread::sendToMemory(shared_ptr<InflightInst> inst_ptr,
...
+        resp->makeResponse();
+
+        Fault f = inst_ptr->staticInst()->completeAcc(resp, inst_ptr.get(),
+                                                      inst_ptr->traceData());
+        // NOTE: SDCPU may do other things for special instruction types, which
+        //       may no longer work correctly with this change.
+
+        delete resp;
+
+        if (f != NoFault) {
```
powerjg: Should we have a panic here?

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -1380,35 +1386,61 @@ SDCPUThread::sendToMemory(shared_ptr<InflightInst> inst_ptr,
...
-        assert(sreq->high && sreq->low);
-        if (write) {
+    if (write) {
+        PacketPtr resp = Packet::createWrite(sreq ? sreq->main : req);
+        if (resp->hasData()) {
+            resp->dataStatic(data.get());
+        }
+        resp->makeResponse();
+
+        Fault f = inst_ptr->staticInst()->completeAcc(resp, inst_ptr.get(),
```
powerjg: When does O3 call this? I know that stores "happen" after commits in the O3 CPU. You can see this if you use O3PipeView. (I can show you later, if you'd like.)

animalvgamer:
Isn't the pipeline viewer basically bufferdump?

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: looks like O3 ignores the return value of completeAcc() altogether :T

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: From further inspection, it actually looks like O3 doesn't call completeAcc() on stores, except in the case of StoreConditionals... funnily enough, we call completeAcc() on all stores, and I haven't made sure to handle the case for StoreConditionals since I was assuming we weren't going to need that case for the paper deadline

animalvgamer: looking at the exec-ns generated implementations, the store instructions basically all don't have any code in completeAcc except return NoFault;

src/cpu/flexcpu/simple_dataflow_thread.cc
powerjg: My thought was that I've been assuming this never faults. I like to have an assert everywhere I make an assumption. But you're right, I *think* we'd handle a fault correctly in this case.

src/cpu/flexcpu/simple_dataflow_thread.cc
powerjg: Interesting. My thought with the pipe view is that I want the trace to have the right timestamp for store completion. However, now that I think more about it, pipeview is it's own special trace, which we haven't implemented yet. Ignore this comment :).

PR #51, opened 2018-10-09
cpu-flex: in-order/in-order-superscalar parameters

animalvgamer:
Change-Id: I5a557a668300efae70b2ed8a839b6a5df62247ab


animalvgamer:
Yep adding constraints seems to be consistently easier than taking them away.

animalvgamer:
As long as the semantic is straightforward anyway

PR #52, opened 2018-11-05
Initial verilator gem5 support

powerjg:
Change-Id: If955751e31360ab11b96b64e9693dc46931ab645


src/verilator/IMem.hh
```diff
@@ -0,0 +1,17 @@
+#ifndef __VERILATOR_IMEM__HH__
+#define __VERILATOR_IMEM__HH__
+
+#include "mem/mem_object.hh"
+
+//verilator file should be here
+
+class IMem
+{
+        protected:
```
powerjg: You should be using 4 spaces for tabs. See the style guideline (http://gem5.org/Coding_Style)

src/verilator/ITop.hh
```diff
@@ -0,0 +1,22 @@
...
+#define __VERILATOR_ITOP__HH__
+
+#include <cstring>
+
+#include "include/VTop.h"
+#include "include/VTop__Dpi.h"
+#include "include/verilated.h"
+#include "include/verilator.h"
+#include "sim/sim_object.hh"
+
```
powerjg: These classes should have some documentation. Doxygen would be best.

src/verilator/README.md
```diff
@@ -0,0 +1,26 @@
+Basic Verilator Wrapper
+* ITop is intended as a wrapper for the top level module of your design
+* IMem is intended as a wrapper for a blackbox verilog memory module
+	-You should use IMem to forward all memory requests to gem5 rather than the c++ model
+* Verilator Object is a basic implementation of ITop for the riscv sodor 1 stage core
+* IMem is currently a work in progrress
+
+Building
+* First you must generate your verilator c++ models, and include them as needed
+	-add flags to the SConscript to modify the compiler flags and include wha tyou need (.i.e. env.Append(CCFLAGS='-I ./<directory>'))
```
powerjg: I don't understand why this is needed. I think there's a better scons solution here... But let's not worry about that this week.

src/verilator/VerilatorObject.py
```diff
@@ -0,0 +1,14 @@
+from m5.params import *
+from m5.SimObject import SimObject
+
+class VerilatorObject(SimObject):
+    type = 'VerilatorObject'
+    cxx_header = "verilator/verilator_object.hh"
+
+    cycles = Param.Int(0, "Maximum number of cycles to simulate,
```
powerjg: I don't think you need this option here. Instead, this is a parameter to m5.simulate() in the runscript.

src/verilator/VerilatorObject.py
```diff
@@ -0,0 +1,14 @@
...
+            0 means unlimited")
+    latency = Param.Latency("Clock period of device under test")
```
powerjg: A better name would be "clockPeriod"...

Actually, if you inherited from "ClockedObject" instead of SimObject, this would come for free (you wouldn't need a parameter). I think that's a better idea.

src/verilator/VerilatorObject.py
```diff
@@ -0,0 +1,14 @@
...
+    stages = Param.Int(1, "Number of stages in the device under test")
```
powerjg: What is a stage in this context?

src/verilator/VerilatorObject.py
```diff
@@ -0,0 +1,14 @@
...
+    startTime = Param.Int(0, "Time to start simulation")
```
powerjg: This should be of type Tick?

src/verilator/VerilatorObject.py
```diff
@@ -0,0 +1,14 @@
...
+    memData = Param.String("Path to data to be loaded in device memory")
+    name = Param.String("VerilatorObject","Object Name")
```
powerjg: The name is populated automatically based on the python variable name in the runscript. This is overloading that automatic variable and I don't think it's needed.

src/verilator/verilator.py
```diff
@@ -0,0 +1,17 @@
+import m5
```
powerjg: This file should be in configs/examples

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+
+VerilatorObject*
+VerilatorObjectParams::create()
+{
+    VerilatorObject * vo = new VerilatorObject(this);
+    return vo;
+}
+
+void
+VerilatorObject::processEvent()
```
powerjg: Let's call this updateCycle() or something like that to make it clear what this function does.

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
+#include "base/logging.hh"
+#include "sim/sim_exit.hh"
+#include "verilator_object.hh"
+
+VerilatorObject::VerilatorObject(VerilatorObjectParams *params) :
+    SimObject(params),
+    event([this]{processEvent();}, params->name),
+    maxCycles(params->cycles),
+    latency(params->latency),
+   designStages(params->stages),
```
powerjg:
```suggestion
    designStages(params->stages),
```

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+   start(params->startTime),
```
powerjg:
```suggestion
    start(params->startTime),
```

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+    loadMem(params->memData.c_str()),
+   objName(params->name)
```
powerjg:
```suggestion
    objName(params->name)
```

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+{
+    VerilatorObject * vo = new VerilatorObject(this);
+    return vo;
+}
+
+void
+VerilatorObject::processEvent()
+{
+    //has verilator finished running?
+    if (dtm->done() && dut.io_success && Verilated::gotFinish()){
```
powerjg: Hopefully we can drop this dtm stuff.

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+        inform("Simulation has Completed\n");
```
powerjg: This would be better as a DPRINTF. In fact, we also need some DPRINTFs in other places, too.

src/verilator/verilator_object.cc
powerjg: `dut.io_success` is specific to the sodor CPU. I think this file shouldn't have any model-specific stuff in it. Verilated::gotFinish() is good here since it's available for all verilated files.

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    //run the device under test here through verilator
+    dut.clock = 0;
+    dut.eval();
+    dut.clock = 1;
+    dut.eval();
+
+    //check if we should stop
+    cyclesPassed += 1;
```
powerjg: This should be a stat.

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+    if (maxCycles != 0 && cyclesPassed == maxCycles){
+        inform("Simulation Timed Out\n");
+        exitSimLoop("Done Simulating", dtm->exit_code());
+    }
+
+    schedule(event, curTick() + latency);
```
powerjg: If this was a `ClockedObject` instead of a `SimObject` (which, BTW, ClockedObject is a sub class of SimObject) then you could just do the following:

```
     schedule(event, nextCycle())
```

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+        dut.eval();
+        //done reseting
+        dut.reset = 0;
+    }
+}
+
+void
+VerilatorObject::startup()
+{
+    reset(designStages);
```
powerjg: Ah! I see what "stages" is used for, now. That should be clearer in the description above.

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,92 @@
...
+        //set reset signal and starting clock signal
+        dut.reset = 1;
+        dut.clock = 0;
+        //run verilator for this state
+        dut.eval();
+        //run verilator for rising edge state
+        dut.clock = 1;
+        dut.eval();
+        //done reseting
+        dut.reset = 0;
```
powerjg: I think this can be outside of the for loop. It's just set back to 1 at the top of the for loop until the end.

In fact, you could move the first reset outside of the for loop, too.

src/verilator/verilator_object.hh
```diff
@@ -0,0 +1,33 @@
+
+#ifndef __VERILATOR_VERILATOR_OBJECT_HHH__
+#define __VERILATOR_VERILATOR_OBJECT_HHH__
+
+#include <fesvr/dtm.h>
```
powerjg: I hope we can drop this.

src/verilator/verilator_object.hh
```diff
@@ -0,0 +1,33 @@
+
+#ifndef __VERILATOR_VERILATOR_OBJECT_HHH__
```
powerjg: Extra H
```suggestion
#ifndef __VERILATOR_VERILATOR_OBJECT_HH__
```

src/verilator/verilator_object.hh
```diff
@@ -0,0 +1,33 @@
...
+#define __VERILATOR_VERILATOR_OBJECT_HHH__
```
powerjg:
```suggestion
#define __VERILATOR_VERILATOR_OBJECT_HH__
```

src/verilator/verilator_object.hh
```diff
@@ -0,0 +1,33 @@
...
+        int designStages;
+        int start;
+        const char * loadMem;
+        std::string const objName;
+        int cyclesPassed;
+
+    public:
+        VerilatorObject(VerilatorObjectParams *p);
+        ~VerilatorObject();
+        void reset(int);
```
powerjg: `int` what? Use variable names in the header file.
```suggestion
        void reset(int reset_cycles);
```

src/verilator/verilator_object.hh
```diff
@@ -0,0 +1,33 @@
...
+
+        dtm_t * dtm;
+        std::vector<std::string> toDtm;
+        EventFunctionWrapper event;
+        int maxCycles;
+        Tick latency;
+        int designStages;
+        int start;
+        const char * loadMem;
+        std::string const objName;
```
powerjg: Not needed. I see how this is a confusing point in my book. I'll try to fix that sometime.

src/verilator/verilator_object.hh
```diff
@@ -0,0 +1,33 @@
...
+#include "sim/sim_object.hh"
+
+class VerilatorObject : public ITop, public SimObject
+{
+    private:
+        void processEvent();
+
+        dtm_t * dtm;
+        std::vector<std::string> toDtm;
+        EventFunctionWrapper event;
```
powerjg: Better name?
```suggestion
        EventFunctionWrapper cycleEvent;
```

src/verilator/verilator_object.hh
```diff
@@ -0,0 +1,33 @@
...
+        int maxCycles;
+        Tick latency;
+        int designStages;
+        int start;
```
powerjg: Is this used? I think it should be a tick anyway.

powerjg:
@nganjehloo: have you had a chance to look over these comments. We need to have something working ASAP so we can make a poster about it :).

nganjehloo:
Hi Jason,

Yup I've been looking through them. I've been wrestling with getting the
memory part working atm.

src/verilator/verilator_object.cc
```diff
@@ -0,0 +1,260 @@
...
+    //ofset address based on data size/type
+    data_req->setPaddr(data_req->getPaddr() + j);
+}
+
+void
+VerilatorObject::sendData(const RequestPtr &req, uint8_t *data, bool read)
+{
+    PacketPtr pkt = buildPacket(req, read);
+    pkt->setData(data);
+    pkt->allocate();
```
powerjg: I don't think you want both setData and allocate. You want one or the other.

Also, when is data deleted? (This is why it might be easier to use allocate() than setData.)

powerjg:
```
pkt->allocate()
pkt->setData(data)
delete[] data;
```

PR #53, opened 2018-11-06
cpu-flex: Simple thread-driven commit-time interrupts.

animalvgamer:
Change-Id: I3325d2e598db517c510f34c27f2c5808b8397774

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -319,6 +319,7 @@ SDCPUThread::commitAllAvailableInstructions()
     shared_ptr<InflightInst> inst_ptr;
 
     bool buffer_shrunk = false;
+    bool interrupts_checked = false;
     while (!inflightInsts.empty()
         && (canCommit(inst_ptr = inflightInsts.front())
             || inst_ptr->isSquashed())) {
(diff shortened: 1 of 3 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -906,6 +940,32 @@ SDCPUThread::onIssueAccessed(weak_ptr<InflightInst> inst)
...
+        curMacroOp = StaticInst::nullStaticInstPtr;
+    }
+
+    const bool interrupt_detected = _cpuPtr->checkInterrupts(this);
+    bool need_advance = true;
+    if (interrupt_detected) {
+        DPRINTF(SDCPUThreadEvent, "Interrupt detected at issue time, clearing "
+                                  "buffer as much as possible.\n");
+        need_advance = false;
+        for (auto& other_inst_ptr : inflightInsts) {
```
powerjg: What happens if the interrupt is handled when there are some instructions on the inflight_inst buffer. I'm worried that it won't be 'consistent' since gem5's interrupt handlers will assume that there are no active instructions in flight (I think this is true). I think we should start with just checking at commit time.

animalvgamer: The interrupt isn't handled at that time, it's checked during issue, just to stop the front-end from continuing beyond what is necessary if an interrupt is detected. The actual handling of the interrupt is delayed until commit time, specifically at the first commit of an operation that is not marked isDelayedCommit().

PR #57, opened 2018-12-20
For upstream/memory tests

akshephard:
tests: Converted simple memory test to new test framework


tests/gem5/memory/simple-run.py
```diff
@@ -0,0 +1,103 @@
...
+parser.add_argument('--latency', default=None)
+parser.add_argument('--latency_var', default=None)
+
+args = parser.parse_args()
+
+# both traffic generator and communication monitor are only available
+# if we have protobuf support, so potentially skip this test
+# require_sim_object("TrafficGen")
+# require_sim_object("CommMonitor")
+# This needs to be fixed in the new infrastructure
```
powerjg: I think this needs to be addressed before upstreaming.

What we want to do is to check to see if those SimObjects exist (I'm not sure where `require_sim_object` lives) and then, if they do not, issue a warning/information. 

This check *can* stay here, but we also need to up it in the testing script. We want this test to be skipped if this requirement isn't met. I'll put a note on where to do that below.

**See below, then read this**
Ok. So, that's not possible/easy. Instead, let's just have this test fail if the requirements aren't met. The only change that needs to be made is to add the checks back in. :)

tests/gem5/memory/test.py
```diff
@@ -0,0 +1,71 @@
...
+        {'latency': '1us', 'name': 'high-latency'},
+        {'bandwidth': '1MB/s', 'name': 'low-bandwidth'},
+        {'latency_var': '100ns', 'name': 'high-var'},
+        ]
+
+for p in simple_mem_params:
+    name = p['name']
+    del p['name']
+    args = ['--' + k + '=' + v for k,v in p.iteritems()]
+
```
powerjg: We *want* to check to see if the SimObjects exist here. However, I don't see any way to do that since gem5 may not even be compiled at this time. So... I guess I'm going to go back and edit my other comment.

tests/gem5/memory/test.py
```diff
@@ -0,0 +1,71 @@
...
+    gem5_verify_config(
+        name='simple_mem_' + name,
+        verifiers=(), # No need for verfiers this will return non-zero on fail
+        config=joinpath(getcwd(), 'simple-run.py'),
+        config_args = args,
+        valid_isas=(constants.null_tag,),
+    )
+
+gem5_verify_config(
```
powerjg: Let's add a comment here that this test actually checks performance, not just functional correctness.

tests/gem5/memory/memtest-run.py
```diff
@@ -0,0 +1,84 @@
...
+# run simulation
+# -----------------------
+
+root = Root( full_system = False, system = system )
+root.system.mem_mode = 'timing'
+
+m5.instantiate()
+exit_event = m5.simulate()
+
+
```
powerjg: Let's add a line here that exit(1) if the exit event isn't what it should be (I'm not sure *exactly* what it should be, but you can run the script and see what it is by printing it like the other script.)

tests/gem5/memory/simple-run.py
```diff
@@ -0,0 +1,103 @@
...
+# run simulation
+# -----------------------
+
+root = Root(full_system = False, system = system)
+root.system.mem_mode = 'timing'
+
+m5.instantiate()
+exit_event = m5.simulate(100000000000)
+
+print(exit_event.getCause())
```
powerjg: In addition to printing this, let's also have a check to make sure it is what it should be (run the script to see what it should be). If the cause is not the right value, call sys.exit(1) so that the test will fail.

tests/gem5/memory/memtest-run.py
akshephard: Added this on for-upstream/memory_tests

tests/gem5/memory/simple-run.py
akshephard: Added this on for-upstream/memory_tests

tests/gem5/memory/test.py
akshephard: Added a comment let me know if more detail is required

tests/gem5/memory/simple-run.py
akshephard: I un-commented these checks and the tests fail. I'm not sure how to proceed on this.

powerjg: The problem is likely that `require_sim_object` isn't in the current scope. Try grepping for that function and import the required module.

tests/gem5/verifier.py
```diff
@@ -46,6 +46,7 @@ def _test(self, *args, **kwargs):
 
     def instantiate_test(self, name_pfx):
         name = '-'.join([name_pfx, self.__class__.__name__])
+        print("Name of test is " + name)
```
powerjg: Looks like this was added for debugging?

tests/gem5/memory/memtest-run.py
powerjg: Great! But let's squash this change into the other two changes (as appropriate). Always feel free to rebase for-upstream/ branches. This is the only way that it works with gerrit upstream.

powerjg:
I'll try to take a look soon. @ayaz91: can you take a look in the meantime?

Thanks!

akshephard:
I was just trying to fix the require_sim_object() issue by moving it to helper.py which is in the testlib itself. I'm editing locally on my machine and pushing the changes so I can use my own editor. I'm kind of stuck on this because I've never really dealt with python code with this many files and I'm not sure how the imports are working.

ext/testlib/helper.py
```diff
@@ -102,6 +102,37 @@ def log_output(log_callback, pipe, redirects=tuple()):
...
+
+    Arguments:
+      name -- Name of SimObject (string)
+
+    Keyword arguments:
+      fatal -- Set to True to indicate that the test should fail
+               instead of being skipped.
+    """
+
+    if has_sim_object(name):
```
ayaz91: will have to add has_sim_object() as well in this file.

animalvgamer:
Already merged to upstream

PR #83, opened 2019-01-24
cpu-flex: Added preprocessor checks to silence warnings w/o trace

animalvgamer:
Change-Id: I4d3f51b379973b94819fa82112ed6de173f6bd29

### DIFF
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -406,13 +406,15 @@ SDCPUThread::commitInstruction(std::shared_ptr<InflightInst> inst_ptr)
 void
 SDCPUThread::dumpBuffer()
 {
+    if (!DTRACE(SDCPUBufferDump)) return;
+
     DPRINTF(SDCPUBufferDump, "%d instructions on the buffer:\n",
                              inflightInsts.size());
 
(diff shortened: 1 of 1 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -406,6 +406,7 @@ SDCPUThread::commitInstruction(std::shared_ptr<InflightInst> inst_ptr)
 void
 SDCPUThread::dumpBuffer()
 {
+#if TRACING_ON
```
powerjg: How about using `if(DTRACE(SDCPUBufferDump))`. I like to avoid the preprocessor directives as much as possible. See http://learning.gem5.org/book/part2/debugging.html#_CPPv26DTRACE4Flag for details

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -1312,6 +1314,7 @@ SDCPUThread::populateUses(shared_ptr<InflightInst> inst_ptr)
 {
     const StaticInstPtr static_inst = inst_ptr->staticInst();
 
+#if TRACING_ON
```
powerjg: This shouldn't be necessary. If you're getting errors about uninitialized variable, use `M5_VAR_USED`

animalvgamer:
Made updates

PR #86, opened 2019-01-24
scons: add -Wno-unused-const-variable flag

takekoputa:
Clang generates errors on unused constant variables, which are created from .cc files that are compiled from .py files. The flag '-Wno-unused-const-variable' is set to suppress those errors.


powerjg:
What's the reason for closing this?

If you're getting unused variable warnings (errors), I think the right thing to do is to use "M5_VAR_USED" preprocessor define on the variables. If it's in SimObject code, you might have to find where the code is generated in python and put them there. We can chat more about this later today.

PR #96, opened 2019-02-14
scons: fix disable_partial logic for fast binary

takekoputa:
Partial linking should be disabled on darwin; however, the script
fails to do so when force_lto is set, which results in gem5 building
with fast option fails on macOS. This fix changes disable_partial
logic, which should be True once it's True.


powerjg:
Can we close this and delete the branch if this has been committed upstream?

If not, why has it not been committed upstream?

takekoputa:
I forgot to close this branch. Sorry about that.

PR #98, opened 2019-02-16
Bradley/flexcpu/mem forward

animalvgamer:


src/cpu/flexcpu/stld_forwarder.cc
```diff
@@ -0,0 +1,498 @@
...
+    });
+    // TODO consider holding onto the store for a little longer (e.g. until
+    //      memory response time), so more loads can be forwarded. Pretty sure
+    //      the other CPU model LSQs will hold onto items until response anyway
+    //      but we haven't implemented it yet only to keep the code simple for
+    //      now.
+    inst_ptr->addCommitCallback([this] { commitStore(); });
+}
+
+// anonymous namespace for internal linkage only.
```
powerjg: This should be in the header as part of the class description.

src/cpu/flexcpu/stld_forwarder.hh
```diff
@@ -0,0 +1,150 @@
...
+
+#include <list>
+#include <memory>
+#include <unordered_map>
+
+#include "base/types.hh"
+#include "cpu/flexcpu/inflight_inst.hh"
+
+class StLdForwarder {
+  protected:
```
powerjg: Planning on having this be a parent class?

src/cpu/flexcpu/stld_forwarder.hh
```diff
@@ -0,0 +1,150 @@
...
+     *                 stack (before requestLoad() returns) if the latency for
+     *                 forwarding is configured to be zero. Packet is nullptr
+     *                 if forwarding is not possible.
+     */
+    void requestLoad(const std::shared_ptr<InflightInst>& inst_ptr,
+                     const RequestPtr& req,
+                     const std::function<void(PacketPtr)>& callback);
+};
+
+#endif // __CPU_FLEXCPU_LSQ_HH__
```
powerjg: Probably should have a newline at the end of the file. Most editors are configured this way...

src/cpu/flexcpu/inflight_inst.cc
```diff
@@ -215,6 +215,12 @@ InflightInst::addReadyCallback(function<void()> callback)
     readyCallbacks.push_back(callback);
 }
 
+void
+InflightInst::addRetireCallback(function<void()> callback)
```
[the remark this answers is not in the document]
animalvgamer: Oh that's actually a good call I didn't think of. I'll try it tomorrow.

src/cpu/flexcpu/stld_forwarder.cc
animalvgamer: Why? It's not important to the interface or internal class definition, so I thought it was a good idea to hide it with the function implementation.

src/cpu/flexcpu/stld_forwarder.hh
animalvgamer: I kind of default to protected over private? I did this to every other class in flexcpu/.

src/cpu/flexcpu/stld_forwarder.hh
animalvgamer: will change when I get in

src/cpu/flexcpu/inflight_inst.cc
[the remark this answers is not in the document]
animalvgamer: I'll make this change in a separate commit/PR, so I can test variants for performance impact.

src/cpu/flexcpu/stld_forwarder.cc
animalvgamer: I moved the declaration into the only function where it's being used, so it's perhaps less weird?

PR #99, opened 2019-03-06
Verilator

powerjg:


configs/example/dinocpu.py
```diff
@@ -0,0 +1,104 @@
...
+
+# Set up the binary to load
+system.kernel = binary
+system.system_port = system.mem_ctrl.port
+
+# Create the dinocpu verilator wrapper
+system.dinocpu = VerilatorDinoCPU()
+
+# Create the black box memory wrapper
+system.memblackbox = VerilatorMemBlackBox()
```
powerjg: The DINO CPU probably needs a pointer to this.

configs/example/dinocpu.py
```diff
@@ -0,0 +1,104 @@
...
+
+system.memblackbox.instPort = system.mem_ctrl.port
+system.memblackbox.dataPort = system.mem_ctrl.port
+
+system.dinocpu.latency = '1ns'
```
powerjg: Should this be a clock period instead of latency?

src/verilator/SConscript
```diff
@@ -0,0 +1,7 @@
+Import('*')
```
powerjg: FYI, we need licenses on all of these files. See other files for examples. The copyright is to University Regents of California or something like that. The DArchR/docs repo has the exact wording.

src/verilator/VerilatorDinoCPU.cc
```diff
@@ -0,0 +1,79 @@
+#include "VerilatorDinoCPU.hh"
+#include "base/logging.hh"
+#include "sim/sim_exit.hh"
+
+VerilatorDinoCPU::VerilatorDinoCPU(VerilatorDinoCPUParams *params) :
+    SimObject(params),
+    event([this]{updateCycle();}, params -> name),
+    latency(params -> latency),
```
powerjg: I've never seen spaces around the `->` operator before. I don't think you should do that. It would be like spaces around the `.` operator. `cpu . memory` doesn't look right :)

src/verilator/VerilatorDinoCPU.hh
```diff
@@ -0,0 +1,31 @@
...
+#include "params/VerilatorDinoCPU.hh"
+#include "sim/sim_object.hh"
+
+class VerilatorDinoCPU
+{
+    private:
+        void updateCycle();
+
+        VTop top;
+        VerilatorMemBlackBox verilatorMem;
```
powerjg: Should probably be a pointer.

Also, to get this pointer into this object, you can add a parameter to the SimObject description file that requires VerilatorMemBlackBox object.

src/verilator/VerilatorDinoCPU.py
```diff
@@ -0,0 +1,9 @@
+from m5.params import *
+from m5.SimObject import SimObject
+
+class VerilatorDinoCPU(SimObject):
+    type = 'VerilatorDinoCPU'
+    cxx_header = "verilator/VerilatorDinoCPU.hh"
+
+    latency = Param.Latency("Clock period of device under test")
+    stages = Param.Int(1, "Number of stages in the device under test")
```
powerjg: Here, you probably also want

```
memory_black_box = Param.VerilatorMemBlackBox("The memory to connect between verilator and gem5")
```

src/verilator/verilator.py
```diff
@@ -0,0 +1,17 @@
+import m5
```
powerjg: Should this be in the configs directory?? I think so!

powerjg (commented):
Looks like we're making progress!!! :D

Could you clean up some of the files that you're not using anymore. You can add a new commit that deletes them, that way you can go back and get them from the old commits if you change your mind.

powerjg:
Watch what user you're using to create commits. The last couple are owned by "root". Did you edit/commit inside the docker container? You probably shouldn't do that.

configs/example/dinocpu.py
nganjehloo: I think I init the memobject in dinocpu.py and pass it as a parameter to DinoCPU in the other config file. Does that not work? I'm still a bit iffy on how the config file passes things around.

nganjehloo (dismissed):

powerjg:
Can you add a commit that deletes all of the files you're not using right now? It's hard to review this PR with a bunch of files that I don't know whether they are being used or not.

Also, you should try not to commit things as root (i.e., in your docker container). You should do all interactions with your file system from outside the container and only use the container to build the .v files.

src/SConscript
```diff
@@ -1378,6 +1378,7 @@ if 'opt' in needed_envs:
             LINKFLAGS = Split(ldflags['opt']),
             disable_partial=disable_partial)
 
+
```
powerjg: Revert this change.

src/verilator/VerilatorDinoCPU.cc
```diff
@@ -0,0 +1,114 @@
+/*# Copyright (c) 2019 The Regents of the University of California
```
powerjg: this should be named verilator_dino_cpu.cc

src/verilator/VerilatorDinoCPU.hh
```diff
@@ -0,0 +1,59 @@
+/*# Copyright (c) 2019 The Regents of the University of California
```
powerjg: Ditto on naming. Underscores, not camel case

src/verilator/VerilatorDinoCPU.hh
```diff
@@ -0,0 +1,59 @@
...
+
+#include "VTop.h"
+#include "VerilatorMemBlackBox.hh"
+#include "params/VerilatorDinoCPU.hh"
+#include "sim/clocked_object.hh"
+
+class VerilatorDinoCPU : public ClockedObject
+{
+    private:
+        void updateCycle();
```
powerjg: Comments would be nice.

src/verilator/VerilatorDinoCPU.hh
```diff
@@ -0,0 +1,59 @@
...
+
+
+#ifndef __VERILATOR_VERILATOR_OBJECT_HH__
+#define __VERILATOR_VERILATOR_OBJECT_HH__
+
+#define VM_TRACE 0
+#define VL_THREADED 0
+
+#include "VTop.h"
+#include "VerilatorMemBlackBox.hh"
```
powerjg: Should have full path for gem5 headers (e.g., verilator/VerilatorMemBlackBox.hh) Except the file should be named with underscores ;).

src/verilator/VerilatorDinoCPU.py
```diff
@@ -0,0 +1,42 @@
...
+
+from m5.params import *
+from VerilatorMemBlackBox import VerilatorMemBlackBox
+from ClockedObject import ClockedObject
+
+class VerilatorDinoCPU(ClockedObject):
+    type = 'VerilatorDinoCPU'
+    cxx_header = "verilator/VerilatorDinoCPU.hh"
+
+    verilatorMem = Param.VerilatorMemBlackBox(NULL, "Reference to "
```
powerjg: Underscores, not camel case.

src/verilator/VerilatorMemBlackBox.hh
```diff
@@ -0,0 +1,83 @@
...
+#define VM_TRACE 0
+#define VL_THREADED 0
+
+#include "VTop_DualPortedMemoryBlackBox.h"
+#include "mem/mem_object.hh"
+#include "params/VerilatorMemBlackBox.hh"
+
+class VerilatorMemBlackBox: public MemObject
+{
+    public:
```
powerjg: Two spaces instead of 4

src/verilator/VerilatorMemBlackBox.hh
```diff
@@ -0,0 +1,83 @@
...
+        VTop_DualPortedMemoryBlackBox* blkbox;
```
powerjg: 4 spaces instead of 8.

src/verilator/VerilatorMemBlackBox.hh
```diff
@@ -0,0 +1,83 @@
...
+
+        void doFetch();
+        void doMem();
+
+        class VerilatorMemBlackBoxPort : public MasterPort
+        {
+            private:
```
powerjg: Ditto above

src/verilator/verilator.py
powerjg: Might want to remove this file.

powerjg (commented):
Rather than checking in *all* of the verilator generated files, it would be better to just check in the .v file and a README and a Makefile on how to generate the rest. This is urgent, but something to think about at some point. Also, you'll want to add a .gitignore in the ext/verilator directory to ignore all of these generated files.

src/verilator/VerilatorDinoCPU.hh
```diff
@@ -0,0 +1,59 @@
...
+
+        VerilatorMemBlackBox * verilatorMem;
+        EventFunctionWrapper event;
+        Tick clkperiod;
+        int designStages;
+        int cyclesPassed;
+        VTop top;
+    public:
+        VerilatorDinoCPU(VerilatorDinoCPUParams *p);
+        ~VerilatorDinoCPU();
```
powerjg: BTW, it's this line that should be removed to get it to compile.

src/verilator/README.md
```diff
@@ -0,0 +1,26 @@
+Basic Verilator Wrapper
```
powerjg: Is this file up to date?

src/verilator/VerilatorDinoCPU.py
```diff
@@ -0,0 +1,42 @@
...
+
+from m5.params import *
+from VerilatorMemBlackBox import VerilatorMemBlackBox
+from ClockedObject import ClockedObject
+
+class VerilatorDinoCPU(ClockedObject):
+    type = 'VerilatorDinoCPU'
+    cxx_header = "verilator/verilator_dino_cpu.hh"
+
+    verilatorMem = Param.VerilatorMemBlackBox("Reference to "
```
powerjg: Underscores (`verilator_mem`) is the correct style.

src/verilator/VerilatorDinoCPU.py
```diff
@@ -0,0 +1,42 @@
...
+        "verilator memory interface")
+    clkperiod = Param.Latency("Clock period of device under test")
```
powerjg: Shouldn't be needed since this is a clocked object.

src/verilator/VerilatorDinoCPU.py
```diff
@@ -0,0 +1,42 @@
...
+    stages = Param.Int(1, "Number of stages in the device under test")
```
powerjg: Is this used?

src/verilator/verilator_dino_cpu.cc
```diff
@@ -0,0 +1,115 @@
...
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+# Authors: Nima Ganjehloo
+*/
+
+#include "VTop_DualPortedMemory.h"
+#include "VTop_DualPortedMemoryBlackBox.h"
+#include "VTop_Top.h"
```
powerjg: Are all of these headers required?

This is where we need to improve the interface :). We would rather have the slimmest interface possible.

src/verilator/verilator_dino_cpu.cc
```diff
@@ -0,0 +1,115 @@
...
+    void* ptr = aligned_alloc(128, sizeof(VerilatorDinoCPU));
+    return new(ptr) VerilatorDinoCPU(this);
+}
+void
+VerilatorDinoCPU::updateCycle()
+{
+
+    DPRINTF(Verilator, "\nTICKING\n");
+    //has verilator finished running?
+    if (Verilated::gotFinish()|| cyclesPassed >= 10){
```
powerjg: Shouldn't need the `cyclesPassed` thing. You can call `simulate(10000)` in the python config file to only simulate 10,000 ticks (10 ns).

src/verilator/verilator_dino_cpu.cc
```diff
@@ -0,0 +1,115 @@
...
+        inform("Simulation has Completed\n");
+        exitSimLoop("Done Simulating", 1) ;
```
powerjg: We may not want to exit the sim loop here, but just stop scheduling the next cycle.

src/verilator/verilator_dino_cpu.cc
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    DPRINTF(Verilator, "INSTRUCTION FETCH?\n");
+
+    verilatorMem->doFetch();
+    verilatorMem->doMem();
+    //run the device under test here through verilator
+    top.clock = 0;
```
powerjg: I think we want to factor this out into a general verilator interface file. This isn't specific to DinoCPU, all verilated objects would want to do this.

src/verilator/verilator_dino_cpu.cc
```diff
@@ -0,0 +1,115 @@
...
+    top.eval();
+    top.clock = 1;
+    top.eval();
+
+    cyclesPassed += 1;
+
+    schedule(event, nextCycle());
```
powerjg: `if (!Verilator::getFinished()) schedule(...)`

src/verilator/verilator_dino_cpu.cc
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+void
+VerilatorDinoCPU::reset(int resetCycles)
```
powerjg: Factor out?

powerjg (commented):
Some comments below. Nothing big :). This is super awesome that you got things working!!!

src/verilator/README.md
nganjehloo: nope will update asap

src/SConscript
nganjehloo: Sorry I'm a bit confused. Revert what?

src/verilator/VerilatorDinoCPU.hh
nganjehloo: For some reason this caused the style checker to freak out. It would keep trying to sort my includes, and would tell me it successfully did, but then when I tried to commit again it would complain despite already being sorted correctly according to the checker :/

src/verilator/VerilatorDinoCPU.py
nganjehloo: It is. I was hoping to test out a 5 or 3 stage cpu at some point. Right now its only used for resetting the cpu.

src/verilator/verilator_dino_cpu.cc
nganjehloo: Yeah unfortunately I was getting a lot of errors. I might be able to import VTop__Allcls.h which would automatically have what I need based on the verilog design. Ill try that out.

src/verilator/verilator_dino_cpu.cc
nganjehloo: Will gem5 automatically stop if there are no events scheduled?

src/verilator/verilator_dino_cpu.cc
nganjehloo: Yup. I was thinking that this seperate routine should schedule an updateCycle event after it does what it needs too. I'll change it like that for the time being.

src/verilator/verilator_dino_cpu.cc
nganjehloo: Factor out as in shouldn't be part of the DinoCPU? LIke put into a testbench class of some sort?

src/SConscript
powerjg: There's an extra newline, according to the diff here.

src/verilator/verilator_dino_cpu.cc
powerjg: Yes.

src/verilator/verilator_dino_cpu.cc
powerjg: Resetting is something you might want to do for all verilated models, not just the DINO CPU. So, this looks like something that should be in the general interface, not in the DINO CPU specifically.

nganjehloo: Ahh gotcha.

ext/verilator/DualPortedMemoryBlackBox.v
```diff
@@ -0,0 +1,33 @@
+module DualPortedMemoryBlackBox(
```
powerjg: This file needs a copyright statement.

ext/verilator/SConscript
```diff
@@ -0,0 +1,14 @@
+Import('main')
```
powerjg: Copyright.

powerjg (commented):
You probably don't want to have the libVTop__ALL.a file checked in.

src/SConscript
```diff
@@ -242,10 +242,12 @@ def blobToCpp(data, symbol, cpp_code, hpp_code=None, namespace=None):
     if hpp_code is not None:
         cpp_code('''\
 #include "blobs/{}.hh"
+
```
powerjg: Still some changes in this file that should be dropped. 

BTW, feel free to rebase this branch. No one else is using is IIRC.

src/verilator/VerilatorDinoCPU.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+from m5.params import *
+from VerilatorMemBlackBox import VerilatorMemBlackBox
+from ClockedObject import ClockedObject
+
+class VerilatorDinoCPU(ClockedObject):
+  type = 'VerilatorDinoCPU'
+  cxx_header = "verilator/verilator_dino_cpu.hh"
+  stages = Param.Int(1, "Number of stages in the device under test")
```
powerjg: What is this meant to be used for. I can tell you have some kind of vision here, but I'm not sure what it is :). Maybe we can talk about it in the Thursday meeting or next time I see you in the lab.

powerjg (commented):
A couple more comments. I didn't mean to submit the other one...

src/verilator/verilator_dino_cpu.cc
animalvgamer: unsigned int?

src/verilator/dpi_manager.cc
```diff
@@ -45,11 +45,28 @@ VerilatorDinoCPU * dinoCPU = nullptr;
...
+  if (imem_address == 0x80000000 ){
+    DPRINTF(Verilator, "DPI INST FETCH ILLEGAL INST EXCEPTION\n");
+    return 0x00c0006f;//jal x0, 12; effectively halt the processor
+  }else if ( imem_address == 0x80000004 ){
+    DPRINTF(Verilator, "DPI INST FETCH TRAPPING FOR ECALL\n");
+    return 0x0080006f;//jal x0, 8; effectively halt the processor
+  }else if ( imem_address == 0x80000008 ){
+    DPRINTF(Verilator, "DPI INST FETCH TRAPPING FOR EBREAK\n");
+    return 0x0040006f;//jal x0, 4; effectively halt the processor
+  }else if ( imem_address > 0x80000008 ){
```
ayaz91: Can your instructions never be mapped beyond this address?

nganjehloo: they easily can but the processor in question is 32 bit so I was intending for the remaining bytes to be reserved for trap/exception handlers. I currently don't redirect the cpu to such handlers because I don't have any that I've written or available to me so I just stall the processor if a trap or exception is raised. Any ideas on how to handle this better?

powerjg:
When's the next time you're going to be in the office? Let's go through the recent updates in person :).

nganjehloo:
I should be there in 1.5 hrs.

powerjg:
Lol. I didn't expect it would be today. I'm not planning on coming in. Will
you be around tomorrow?

nganjehloo:
Oh right. It's a holiday lol. Yeah i will be in tomorrow as well.

powerjg:
You may want to check out https://github.com/powerjg/gem5/blob/devel/accel which has code for using devices/drivers in SE mode.

src/verilator/nvdla/NVDLA.py
```diff
@@ -33,7 +33,8 @@
 class NVDLAWrapper(DrivenObject):
   type = 'NVDLAWrapper'
   cxx_header = "verilator/nvdla/nvdla_wrapper.hh"
-  do_trace = Param.Bool(True, "Use the specified trace file?")
   buf_clear_cycles = Param.Int(4096,
     "Number of cycles to clear hardware buffers")
-  trace_file = Param.String("input.txn", "Name of trace file to run test" )
+  nvdla_sys = Param.System(Parent.any, "Pointer to nvdla system")
+  axi_2_gem5 = Param.MemObject(Parent.any,
```
powerjg: Will this really be a parent? I suggest not using a proxy here.

src/verilator/nvdla/axi_to_mem.hh
```diff
@@ -0,0 +1,85 @@
...
+    public VerilatorMemBlackBox::VerilatorMemBlackBoxPort
+    {
+      public:
+
+         AXIToMemPort(const std::string& name,
+                    AXIToMem *owner) :
+                    AXIToMemPort(name, owner)
+                { }
+
+        vvoid sendTimingPacket(PacketPtr pkt);
```
powerjg: typo.

src/verilator/nvdla/NVDLA.py
nganjehloo: Yeah I wasn't really sure what would go here. based off the Proxy Factory code I figured any would go down the tree and eventually hit what I was asking for. Would the system i passed have a way of getting the memobject? I couldn't find any getters in the class declaration

powerjg: No.. you would probably want to do the following in the config file:

```
system.nvdla = NVDLAWrapper()
system.nvdla.axi_2_gem5 = AXIWrapper()
```
I'm not exactly sure what the `AXIWrapper` is here or if you have that done already. But basically, you don't think you want the `nvdla` to be a child of it's memory "port". Instead you want the memory port to be a child of the `nvdla`.

nganjehloo: I see. I think I fixed it.

PR #102, opened 2019-03-15
cpu-flex: Added alternative constraint to limit insts instead of ops.

animalvgamer:
Change-Id: I81074541d5f028c09541ce09082e345f442349a3

### DIFF
--- a/src/cpu/flexcpu/SimpleDataflowCPU.py
+++ b/src/cpu/flexcpu/SimpleDataflowCPU.py
@@ -83,15 +83,27 @@ class SimpleDataflowCPU(BaseCPU):
     in_order_execute = Param.Bool(False, "Serialize all instruction "
                                          "execution.")
 
-    instruction_buffer_size = Param.Unsigned(0, "Size of the dynamic "
-                                             "instruction buffer. This buffer "
-                                             "is used for maintaining the "
-                                             "commit order of instructions. "
-                                             "Limiting this limits the number "
-                                             "of instructions that can be "
-                                             "handled at once before commit "
-                                             "(out of order). 0 implies an "
-                                             "infinitely large buffer.")
+    op_buffer_size = Param.Unsigned(0, "Size of the InflightInst buffer. This "
+                                       "buffer stores decoded (possibly "
+                                       "microcoded) instructions used for "
+                                       "maintaining the commit order of "
+                                       "instructions and microops. Limiting "
+                                       "this limits the number of "
+                                       "instructions AND microops that can be "
+                                       "handled at once before commit (out of "
+                                       "order). 0 implies an infinitely large "
+                                       "buffer.")
+
+    max_instruction_window = Param.Unsigned(0, "Normally, the instruction "
+                                            "buffer is bound by the number of "
+                                            "microops that instructions are "
+                                            "decoded into (1 for instructions "
+                                            "that are not microcoded). This "
+                                            "parameter allows for a separate "
+                                            "constraint to be applied just to "
+                                            "the count of ISA instructions, "
+                                            "ignoring any effects of "
+                                            "microcode.")
 
     clocked_issue = Param.Bool(Self.clocked_cpu, "Ties requests for issuing "
                                                  "instructions to the clock")
--- a/src/cpu/flexcpu/simple_dataflow_thread.cc
+++ b/src/cpu/flexcpu/simple_dataflow_thread.cc
@@ -52,7 +52,8 @@ SDCPUThread::SDCPUThread(SimpleDataflowCPU* cpu_, ThreadID tid_,
                     TheISA::ISA* isa_, bool use_kernel_stats_,
                     unsigned branch_pred_max_depth, unsigned fetch_buf_size,
                     bool in_order_begin_exec, bool in_order_exec,
-                    unsigned inflight_insts_size, bool strict_ser,
+                    unsigned inflight_insts_size,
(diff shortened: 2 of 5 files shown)


src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -985,6 +1022,8 @@ SDCPUThread::onInstDataFetched(weak_ptr<InflightInst> inst,
 
             decode_result = curMacroOp->fetchMicroop(pc.microPC());
 
+            if (!decode_result->isFirstMicroop()) --numInstsInWindow;
```
powerjg: So we subtract from `numInstsInWindow` for *all* microops except the first? This confuses me.

Maybe I don't understand what `numInstsInWindow` is.

src/cpu/flexcpu/simple_dataflow_thread.hh
```diff
@@ -132,6 +132,7 @@ class SDCPUThread : public ThreadContext
     const bool inOrderExecute;
 
     const unsigned inflightInstsMaxSize;
```
powerjg: This would be better named "inflightOpsMaxSize" right? Maybe you should do that.

src/cpu/flexcpu/simple_dataflow_thread.cc
```diff
@@ -203,17 +212,38 @@ SDCPUThread::advanceInst(TheISA::PCState next_pc)
...
 
+        DPRINTF(SDCPUThreadEvent, "Buffer size increased to %d\n",
+                                  inflightInsts.size());
+
         issueInstruction(inst_ptr);
 
         return;
     }
 
+    advanceInstNeedsRetry = static_cast<bool>(maxInstWindow)
```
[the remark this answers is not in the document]
animalvgamer: yea it's just checking for nonzero, I just thought the explicit cast would be better than an implicit one? It is technically completely redundant.

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: Sometimes a microcoded instruction will branch to a new pc within the same macroop which causes a fetch to occur, even though it never completed the prior one. This check is to undo the increment which happened in advanceInst in that case.

src/cpu/flexcpu/simple_dataflow_thread.hh
animalvgamer: I thought about it, but I figured it would be better to keep the name for the inflightInsts buffer, this variable, and the class name InflightInst consistent? I'm not sure if all the things should be changed :L

src/cpu/flexcpu/SimpleDataflowCPU.py
```diff
@@ -83,15 +83,27 @@ class SimpleDataflowCPU(BaseCPU):
...
+                                       "microcoded) instructions used for "
+                                       "maintaining the commit order of "
+                                       "instructions and microops. Limiting "
+                                       "this limits the number of "
+                                       "instructions AND microops that can be "
+                                       "handled at once before commit (out of "
+                                       "order). 0 implies an infinitely large "
+                                       "buffer.")
+
+    max_instruction_window = Param.Unsigned(0, "Normally, the instruction "
```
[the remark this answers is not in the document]
animalvgamer: They're two independently configurable constraints, and specifying both at the same time can have different impacts depending on the ratio of microcoded instructions to non-microcoded instructions in the workload.

src/cpu/flexcpu/simple_dataflow_thread.hh
animalvgamer: Or if not all, which parts.

animalvgamer:
I didn't realize this had already been reviewed, as I had kind of forgotten about the PR lol

src/cpu/flexcpu/simple_dataflow_thread.cc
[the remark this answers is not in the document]
powerjg: No, I mean it should say `advanceInstNeedsRetry = maxInstWindow > 0`

src/cpu/flexcpu/simple_dataflow_thread.cc
powerjg: Can you add this as a comment? The code is not self explanatory.

src/cpu/flexcpu/simple_dataflow_thread.hh
powerjg: Maybe you should change all of the names, actually.

Or keep it this way and make sure to add documentation to make it clear that "Inst" means `StaticInst`, which is a micro-op (maybe).

src/cpu/flexcpu/simple_dataflow_thread.cc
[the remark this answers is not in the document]
animalvgamer: done

src/cpu/flexcpu/simple_dataflow_thread.cc
animalvgamer: done

src/cpu/flexcpu/simple_dataflow_thread.hh
animalvgamer: Added documentation to the InflightInst header. I'm not sure if there is a cleaner solution, but I just hope that the documentation that's associated with the parameters themselves is sufficient.

animalvgamer:
I can merge this whenever if this is good.

PR #123, opened 2019-05-07
cpu-flex: Added Branch Prediction functionality to FlexCPU

animalvgamer:
cpu-flex: Added Branch Prediction functionality (auxiliary changes removed from 569966e)

flexcpu: Modify BP to not fetch if it's wrong

Instead of trying to fetch more instructions when we *know* the branch
predictor is wrong, simply stop the fetch unit. This will improve the
performance of the simulator (from 408s to 103s for Bubblesort). This
slightly changes the simulated system's performance, but by much less
that 1% for Bubblesort, and it's unclear if this change is anything
except random noise.

flexcpu: Add stats (with bp stats, original 2a390ba)

Add stats for branch predictor mis-prediction latency and
the number of instructions fetched incorrectly

cpu-flex: fixed case where bpd was not released if buffer full

cpu-flex: Remove redundant list for branches past prediction limit

Change-Id: I355410665e3047119b1358391eade084bc5b453e

### DIFF
--- a/src/cpu/flexcpu/flexcpu_thread.cc
+++ b/src/cpu/flexcpu/flexcpu_thread.cc
@@ -30,11 +30,13 @@
 
 #include "cpu/flexcpu/flexcpu_thread.hh"
 
+#include <algorithm>
 #include <string>
 
 #include "base/intmath.hh"
 #include "base/trace.hh"
 
+#include "debug/FlexCPUBranchPred.hh"
 #include "debug/FlexCPUBufferDump.hh"
 #include "debug/FlexCPUDeps.hh"
 #include "debug/FlexCPUInstEvent.hh"
@@ -48,6 +50,7 @@ using namespace std;
 FlexCPUThread::FlexCPUThread(FlexCPU* cpu_, ThreadID tid_,
                     System* system_, BaseTLB* itb_, BaseTLB* dtb_,
                     TheISA::ISA* isa_, bool use_kernel_stats_,
+                    unsigned branch_pred_max_depth,
                     unsigned fetch_buf_size, unsigned inflight_insts_size,
                     bool strict_ser):
     memIface(*this),
@@ -59,7 +62,9 @@ FlexCPUThread::FlexCPUThread(FlexCPU* cpu_, ThreadID tid_,
     isa(isa_),
     fetchBuf(vector<uint8_t>(fetch_buf_size)),
     fetchBufMask(~(static_cast<Addr>(fetch_buf_size) - 1)),
-    inflightInstsMaxSize(inflight_insts_size)
+    inflightInstsMaxSize(inflight_insts_size),
+    remainingBranchPredDepth(branch_pred_max_depth ?
+                             branch_pred_max_depth : -1)
 {
     fatal_if(fetch_buf_size % sizeof(TheISA::MachInst) != 0,
              "Fetch buffer size should be multiple of instruction size!");
@@ -70,7 +75,8 @@ FlexCPUThread::FlexCPUThread(FlexCPU* cpu_, ThreadID tid_,
 // Non-fullsystem constructor
 FlexCPUThread::FlexCPUThread(FlexCPU* cpu_, ThreadID tid_,
                     System* system_, Process* process_, BaseTLB* itb_,
-                    BaseTLB* dtb_, TheISA::ISA* isa_, unsigned fetch_buf_size,
+                    BaseTLB* dtb_, TheISA::ISA* isa_,
+                    unsigned branch_pred_max_depth, unsigned fetch_buf_size,
                     unsigned inflight_insts_size, bool strict_ser):
     memIface(*this),
     _cpuPtr(cpu_),
@@ -81,7 +87,9 @@ FlexCPUThread::FlexCPUThread(FlexCPU* cpu_, ThreadID tid_,
     isa(isa_),
     fetchBuf(vector<uint8_t>(fetch_buf_size)),
     fetchBufMask(~(static_cast<Addr>(fetch_buf_size) - 1)),
-    inflightInstsMaxSize(inflight_insts_size)
+    inflightInstsMaxSize(inflight_insts_size),
+    remainingBranchPredDepth(branch_pred_max_depth ?
+                             branch_pred_max_depth : -1)
 {
     fatal_if(fetch_buf_size % sizeof(TheISA::MachInst) != 0,
              "Fetch buffer size should be multiple of instruction size!");
@@ -369,6 +377,11 @@ FlexCPUThread::commitInstruction(InflightInst* const inst_ptr)
 
     inst_ptr->commitToTC();
 
+    if (_cpuPtr->hasBranchPredictor() && inst_ptr->staticInst()->isControl()) {
+        _cpuPtr->getBranchPredictor()->update(inst_ptr->issueSeqNum(),
+                                              threadId());
+    }
+
     if (!inst_ptr->staticInst()->isMicroop() ||
            inst_ptr->staticInst()->isLastMicroop()) {
         numInstsStat++;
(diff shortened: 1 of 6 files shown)


src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -120,18 +128,18 @@ FlexCPUThread::activate()
...
-    }
-
     advanceInstNeedsRetry = static_cast<bool>(inflightInstsMaxSize)
                          && inflightInsts.size() >= inflightInstsMaxSize;
     if (advanceInstNeedsRetry) {
         advanceInstRetryPC = next_pc;
         return;
     }
 
+    if (status() != ThreadContext::Active) {
```
powerjg: Why is this now after the retry? This doesn't make sense to me. If it's not active, why would we fetch?

src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -569,11 +605,73 @@ FlexCPUThread::markFault(shared_ptr<InflightInst> inst_ptr, Fault fault)
...
+    // BPredUnit::predict takes pc by reference, and updates it in-place.
+
+    DPRINTF(FlexCPUBranchPred, "(seq %d) predicted %s (predicted pc: %#x).\n",
+                              inst_ptr->seqNum(),
+                              taken ? "taken" : "not taken",
+                              pc.instAddr());
+    // Count stats using the predict return value?
+
+    StaticInstPtr static_inst = inst_ptr->staticInst();
+    if (static_inst->isDirectCtrl() && static_inst->isUncondCtrl()) {
```
powerjg: Is the `&&` right here? It shouldn't be `||`? just checking...

src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -569,11 +605,73 @@ FlexCPUThread::markFault(shared_ptr<InflightInst> inst_ptr, Fault fault)
...
+    } else if ((static_inst->isCall() || static_inst->isReturn()) && !taken) {
+        // Calls are always taken.
+        DPRINTF(FlexCPUBranchPred, "Call/return taken misprediction\n");
+        freeBranchPredDepth();
+        return;
+    }
+
+    inst_ptr->addSquashCallback([this, inst] {
+        const shared_ptr<InflightInst> inst_ptr = inst.lock();
+        if (!inst_ptr || inst_ptr->isCommitted() || inst_ptr->isComplete())
```
powerjg: This doesn't make sense. Shouldn't we be able to use an `assert` here? It seems that if someone calls the squash callback, this instruction *must* be squashed.

src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -569,11 +605,73 @@ FlexCPUThread::markFault(shared_ptr<InflightInst> inst_ptr, Fault fault)
...
+            return;
+
+        freeBranchPredDepth();
+    });
+
+
+    advanceInst(pc);
```
powerjg: Does this take into account the fetch bandwidth limit? I think so but...

src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -652,9 +750,76 @@ FlexCPUThread::onExecutionCompleted(shared_ptr<InflightInst> inst_ptr,
...
-        advanceInst(getNextPC());
+        if (_cpuPtr->hasBranchPredictor()) {
+            // If a branch predictor has been set, then we need to check if
+            // a prediction was made, squash instructions if incorrectly
+            // predicted, and notify the predictor accordingly.
+
+            auto it = find(inflightInsts.begin(), inflightInsts.end(),
+                           inst_ptr);
+
+            if (it == inflightInsts.end()) {
```
powerjg: `panic_if`

src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -652,9 +750,76 @@ FlexCPUThread::onExecutionCompleted(shared_ptr<InflightInst> inst_ptr,
...
+                // Should be unreachable...
+                panic("Did a completion callback on an instruction not in "
+                      "our buffer.");
+            }
+
+            shared_ptr<InflightInst> following_inst =
```
powerjg: How about if next iterator (whatever that looks like in C++) is not the end or `advanceInstNeedsRetry`. Then you don't have to have the null check.

src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -652,9 +750,76 @@ FlexCPUThread::onExecutionCompleted(shared_ptr<InflightInst> inst_ptr,
...
+
+                if (correctPC.pc() == predictedPC.pc()
+                 && correctPC.upc() == predictedPC.upc()) {
+                    // It was a correct prediction
+                    DPRINTF(FlexCPUBranchPred,
+                            "Branch predicted correctly (seq %d)\n",
+                            inst_ptr->seqNum());
+                    // Calculate stat correct predictions?
+                    // Otherwise do nothing because we guessed correctly.
+                    // Just remember to update the branch predictor at commit.
```
powerjg: Why not update the branch predictor now? This could be a reason we have so many incorrect branches...

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: Can't think of why either. It might be an artifact of the rebasing that was done when reordering commits. I'm going to remove these patches from the commit.

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: Pretty sure it's right, since indirectctrl can't be known early as "taken" or "not taken", since a register value used to jump to could be the next instruction, whereas this logic describes an assumption that a compiler wouldn't create an unconditional direct jump to the immediate next instruction.

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: it's not a panic call, it's a decision as to whether or not to freeBranchPredDepth(), based on whether it reached completion before it was squashed or not.

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: I'm not sure how to address this, since:

1) bandwidth constraints are managed by the CPU, so this still goes through that channel.
2) We don't actually have a configurable fetch bandwidth, but an issue bandwidth,  which this instruction will still have to go through.

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: I'd rather leave that as a block, because it might someday become possible, if there exist store instructions which are also branch instructions (since they can be committed and removed from the buffer before they actually have an onExecutionCompleted call)? It's an edge case but it might make it easier to deal with if that comes.

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: I don't like the way the code looks if we separate out the iterator access that way. More lines, and we still have to have the null check again about 7 lines later.

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: We don't know that this branch will even be committed.

We could have like ld r1, 100 -> br r1,16 (mispredict?) ->ld r2, 200 -> br r2, 32 (correct?), where some previous instruction allows r2 to be read before r1. the second branch completes execution first, but we don't know if its "correct" prediction is actually even part of code which should be executed.

src/cpu/flexcpu/flexcpu_thread.cc
animalvgamer: isCommitted() is redundant though, that shouldn't be there.

animalvgamer: also it can't be null since it's held and called in its own callback. I'm going to use a raw pointer here instead

PR #126, opened 2019-05-07
cpu-flex: Added some necessary code for fullsystem

animalvgamer:
These are features that are necessary for fullsystem simulation.
However, no one has fully tested flexcpu for fullsystem yet.

cpu-flex: Added missing initialization code for fullsystem

cpu-flex: Simple thread-driven commit-time interrupts.

Change-Id: I540c331974cb42ef8985e0dcbe87e1e29f57a0e5

### DIFF
--- a/src/cpu/flexcpu/flexcpu_thread.cc
+++ b/src/cpu/flexcpu/flexcpu_thread.cc
@@ -326,6 +326,7 @@ FlexCPUThread::commitAllAvailableInstructions()
     shared_ptr<InflightInst> inst_ptr;
 
     bool buffer_shrunk = false;
+    bool interrupts_checked = false;
     while (!inflightInsts.empty()
         && (canCommit(*(inst_ptr = inflightInsts.front()))
             || inst_ptr->isSquashed())) {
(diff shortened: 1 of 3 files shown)


src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -345,6 +346,17 @@ FlexCPUThread::commitAllAvailableInstructions()
 
         inflightInsts.pop_front();
         buffer_shrunk = true;
+
+        if (!interrupts_checked
```
powerjg: Can you add a comment describing when interrupts are handled? Is this as soon as they are posted no matter what?

src/cpu/flexcpu/flexcpu_thread.cc
```diff
@@ -913,6 +948,32 @@ FlexCPUThread::onIssueAccessed(weak_ptr<InflightInst> inst)
     inst_ptr->traceData(nullptr);
 #endif
 
+    if (inst_ptr->staticInst()->isLastMicroop()) {
+        DPRINTF(FlexCPUThreadEvent, "Releasing MacroOp after decoding final "
+                                    "microop\n");
+        curMacroOp = StaticInst::nullStaticInstPtr;
+    }
+
+    const bool interrupt_detected = _cpuPtr->checkInterrupts(this);
```
powerjg: Why is this needed at issue time?

animalvgamer: This is for optimization in the same way branch prediction ends early if we think we can know earlier that an interrupt occurred. We kill any issue logic when an interrupt occurs so that the CPU spends less time issuing, when an interrupt is going to squash all of it anyway on the next time a commit occurs.

PR #130, opened 2019-05-09
Removed size from simple_dataflow_thread.cc and added override keywor…

Cazamere:
Removed size from simple_dataflow_thread.cc and added override keywords to simple_mem.hh


animalvgamer:
Looks good to me

powerjg:
@animalvgamer Can you merge this? I'm not sure what technique would work best for you.

PR #161, opened 2019-08-24
For reference only

takekoputa:
The code is quite messy now, and the commit messages are not very useful. I'll make different commits for those changes.


powerjg:
Can you make the PR into flexcpu so I can concentrate just on the changes for ARM?

PR #170, opened 2019-10-30
Verifier

powerjg:


tests/gem5/verifier.py
```diff
@@ -91,11 +91,18 @@ def test(self, params):
                             self.test_filename,
                             ignore_regexes=self.ignore_regex,
                             logger=params.log)
+
+        for line in diff:
```
powerjg: I think we probably want to make a subclass of `MatchGoldStandard`. Something like `MatchStats`. This will take as a parameter the stats you want to match (say, `['committedInsts', 'cpu.cycles', ...]`) and check that none of these are in the diff.

BTW, I like the way you're checking the diff. I would have parsed the stats file, but your way is definitely better. :).

powerjg (commented):
Small comment below, but I think it was the direction you were going anyway. Also, make sure to clean up the diff and the commits so that it only has the changes to the tests :).

tests/gem5/verifier.py
```diff
@@ -113,6 +107,31 @@ def _generic_instance_warning(self, kwargs):
                              ' generic %s'
                              ' instead' % MatchGoldStandard.__name__)
 
+class MatchStat(MatchGoldStandard):
+    _stats = []
```
powerjg: No need to declare this as a (static) class variable.

tests/gem5/verifier.py
```diff
@@ -113,6 +107,31 @@ def _generic_instance_warning(self, kwargs):
...
+
+    def __init__(self, stat):
+        self._stats = stats
```
powerjg: I'd use `self.stats`, personally.

tests/gem5/verifier.py
```diff
@@ -113,6 +107,31 @@ def _generic_instance_warning(self, kwargs):
...
+        tempdir = fixtures[constants.tempdir_fixture_name].path
+        self.test_filename = joinpath(tempdir, self.test_filename)
+
+        diff = diff_out_file(self.standard_filename,
+                            self.test_filename,
+                            ignore_regexes=self.ignore_regex,
+                            logger=params.log)
+
+        for line in diff:
+            for stat = self._stats:
```
powerjg: Is this syntax correct? I think you mean `for stat in self._stats:`

tests/gem5/verifier.py
```diff
@@ -113,6 +107,31 @@ def _generic_instance_warning(self, kwargs):
...
+                if stat in line:
+                    self.failed(fixtures)
+                    test.fail('Stats did not match:\n%s\nSee %s for results'
```
powerjg: Do we need to return after this? Or does this call `sys.exit()`?

tests/gem5/cpu_tests/test.py
```diff
@@ -44,14 +44,14 @@
...
 base_path = joinpath(absdirpath(__file__), 'benchmarks', 'bin')
 base_url = 'http://gem5.org/dist/current/gem5/cpu_tests/benchmarks/bin/'
+stats = ['committedInsts', 'cpu.cycles']
 for isa in valid_isas:
     path = joinpath(base_path, isa)
     for workload in workloads:
         ref_path = joinpath(getcwd(), 'ref', workload)
         verifiers = (
-                verifier.MatchStats(ref_path),
+                verifier.MatchStat(stats),
```
powerjg: Perfect!

powerjg (commented):
It looks like the branch needs to be cleaned up a bit still. Maybe a simple rebase on master to remove some of the extraneous changes?

Otherwise, a few small comments below. Generally, this looks exactly like I'd envisioned!

tests/gem5/verifier.py
```diff
@@ -203,4 +226,4 @@ def parse_file(fname):
 def _iterable_regex(regex):
     if isinstance(regex, _re_type) or isinstance(regex, str):
         regex = (regex,)
-    return regex
+    return regex
```
powerjg: Lost a newline :)

tests/gem5/verifier.py
```diff
@@ -106,6 +106,29 @@ def _generic_instance_warning(self, kwargs):
                              ' generic %s'
                              ' instead' % MatchGoldStandard.__name__)
 
+class MatchStat(MatchGoldStandard):
```
powerjg: Let's call this match stats since you can match more than 1

Also, it would be good to add a brief docstring. Maybe something like:

```python
"""
For a list of stats, this verifier checks if any of them have changed. If so, the test fails.
This class does a simple diff like MatchGoldStandard, but then only checks for the stats
in the diff. This allows some stats to change (e.g., host stats) without causing the test to fail.
"""
```

tests/gem5/verifier.py
```diff
@@ -106,6 +106,29 @@ def _generic_instance_warning(self, kwargs):
...
+    def __init__(self, stat):
```
powerjg: Same thing here: "stats" instead of "stat"

powerjg (commented):
Looking close! A couple of notes below :).

1. the commit message should be
```
tests: Add a new verifier for stats

For a list of stats, this verifier checks if any of them have changed. If so, the test fails.
This class does a simple diff like MatchGoldStandard, but then only checks for the stats
in the diff. This allows some stats to change (e.g., host stats) without causing the test to fail.
```
2. The two commits in the branch should be squashed. You can use git rebase -i master and mark the second one as "squash". See https://git-scm.com/book/en/v2/Git-Branching-Rebasing for more info.

tests/gem5/cpu_tests/test.py
```diff
@@ -49,7 +49,7 @@
     for workload in workloads:
         ref_path = joinpath(getcwd(), 'ref', workload)
         verifiers = (
-                verifier.MatchStdout(ref_path),
+                verifier.MatchStats(ref_path),
```
powerjg: Just out of curiosity, does it work to have both verifiers? If so, I think we should have both.

powerjg (commented):
Other than the minor comment below, and rebasing into a single commit, this is ready to go to gerrit. Feel free to push there without me re-reviewing here :)

tests/gem5/cpu_tests/test.py
```diff
@@ -49,7 +49,8 @@
     for workload in workloads:
         ref_path = joinpath(getcwd(), 'ref', workload)
         verifiers = (
-                verifier.MatchStdout(ref_path),
+                verifier.MacthStdout(ref_path),
+                verifier.MatchStats(ref_path),
```
powerjg: Two questions:
1) doesn't MatchStats take in only the list of stats?
2) Shouldn't MatchStats also need the ref_path?

powerjg:
Could you update the commit message to fit the gem5 style. Something like what's below: (note the lines should be <75 chars, I just eyeballed it so double check me :))

```
tests: Add a match stats verifier

This new verifier compares specific stats to a reference. This changeset also updates the the 
CPU tests to use the new verifier to check the ???? stat.

```

PR #177, opened 2020-07-23
Remove master/slave terminology

powerjg:
This is a WIP. I went ahead and created a PR for easy of looking at the changes :)

### DIFF
--- a/src/mem/port.hh
+++ b/src/mem/port.hh
@@ -245,7 +245,8 @@ class RequestPort: public Port, public AtomicRequestProtocol,
 }
 
 [[deprecated]]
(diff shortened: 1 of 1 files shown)


src/mem/port.hh
```diff
@@ -68,6 +68,183 @@ class SlavePort;
...
+    recvRetrySnoopResp() override
+    {
+        panic("%s was not expecting a snoop retry.\n", name());
+    }
+};
+
+}
+
+[[deprecated]]
 class MasterPort : public Port, public AtomicRequestProtocol,
```
powerjg, line 248: Do you think the following would work?

```suggestion
class MasterPort : public RequestPort
{};
```

PR #178, opened 2020-07-27
Req resp

shparekh00:
This branch is about changing the master slave terminology to requestor responder.
Jira Issue: https://gem5.atlassian.net/browse/GEM5-9


src/mem/AddrMapper.py
```diff
@@ -37,19 +37,19 @@
...
 # currently not modified.
 class AddrMapper(SimObject):
     type = 'AddrMapper'
     cxx_header = 'mem/addr_mapper.hh'
     abstract = True
 
     # one port in each direction
-    master = MasterPort("Master port")
-    slave = SlavePort("Slave port")
+    requestor = RequestPort("Request port")
```
powerjg, line 51: Let's update this documentation while we're here... (sorry for pointing this out a bit late). Something like what's below.

```suggestion
    requestor = RequestPort("Port which forwards the original request to the 'memory side.' The packet being sent as the request will have the new address based on the configured mapping.")
```

src/mem/noncoherent_xbar.cc
```diff
@@ -53,34 +53,35 @@
...
                                                               defaultPortID)));
     }
 
-    // create the slave ports, once again starting at zero
-    for (int i = 0; i < p->port_slave_connection_count; ++i) {
-        std::string portName = csprintf("%s.slave[%d]", name(), i);
-        QueuedSlavePort* bp = new NoncoherentXBarSlavePort(portName, *this, i);
-        slavePorts.push_back(bp);
+    // create the response ports, once again starting at zero
+    for (int i = 0; i < p->port_response_connection_count; ++i) {
```
LakinS, line 80: I was thinking that this might be confusing. Would variables like this with 'port_response' instead of 'response_port' be better as 'port_responder' ?
-Lakin

src/mem/AddrMapper.py
```diff
@@ -37,19 +37,19 @@
...
+    responder = ResponsePort("Response port")
```
powerjg, line 52: Maybe we should call this receiver? 

```suggestion
    receiver = ResponsePort("Port that receives requests and translates them before forwarding out of the requestor port.")
```

src/mem/AddrMapper.py
```diff
@@ -37,19 +37,19 @@
 from m5.SimObject import SimObject
 
 # An address mapper changes the packet addresses in going from the
```
powerjg, line 39: We should move this comment to be a true python documentation comment:

E.g.,

```
"""An address mapper changes...
"""
```

And it should be under `class AddrMapper`

Finally, it would be good to mention here that you can use the `splice` function to insert this address mapper between two already connected objects.

powerjg (commented):
A couple of comments to chat about.

PR #180, opened 2020-07-28
Deprecate SlavePort and MasterPort Classes and change terminology

shparekh00:
Changing master slave terminology.

https://gem5.atlassian.net/browse/GEM5-9

Change-Id: I5e6e90b2916df4dbfccdaabe97423f377a1f6e3f

### DIFF
--- a/src/mem/port.cc
+++ b/src/mem/port.cc
@@ -50,7 +50,7 @@
 namespace
 {
 
-class DefaultMasterPort : public MasterPort
+class DefaultRequestPort : public RequestPort
 {
   protected:
     [[noreturn]] void
@@ -60,7 +60,7 @@ class DefaultMasterPort : public MasterPort
     }
 
   public:
-    DefaultMasterPort() : MasterPort("default_master_port", nullptr) {}
+    DefaultRequestPort() : RequestPort("default_request_port", nullptr) {}
 
     // Atomic protocol.
     Tick recvAtomicSnoop(PacketPtr) override { blowUp(); }
@@ -75,7 +75,7 @@ class DefaultMasterPort : public MasterPort
     void recvFunctionalSnoop(PacketPtr) override { blowUp(); }
 };
 
-class DefaultSlavePort : public SlavePort
+class DefaultResponsePort : public ResponsePort
 {
   protected:
     [[noreturn]] void
@@ -85,7 +85,7 @@ class DefaultSlavePort : public SlavePort
     }
 
   public:
-    DefaultSlavePort() : SlavePort("default_slave_port", nullptr) {}
+    DefaultResponsePort() : ResponsePort("default_response_port", nullptr) {}
 
     // Atomic protocol.
     Tick recvAtomic(PacketPtr) override { blowUp(); }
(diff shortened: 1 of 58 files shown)


src/mem/port.hh
```diff
@@ -373,52 +383,52 @@ class SlavePort : public Port, public AtomicResponseProtocol,
...
         }
     }
 
   protected:
     /**
-     * Called by the master port to unbind. Should never be called
+     * Called by the requestor port to unbind. Should never be called
      * directly.
      */
     void slaveUnbind();
```
ebrickey, line 425: There were two functions (slaveUnbind() and slaveBind()) we were unsure how to deprecate. In the RequestPort class should this now be RequestPort::requestorUnbind() and in the SlavePort class we'd need SlavePort::slaveUnbind(). Both of which would need to be identical functions with the same contents just a different name that we need to define in the .cc? These were the only two cases we've seen so far where the old terminology effected function names in this class.

powerjg: You should be able to just change the names of these functions. This shouldn't be an exposed API, and we should be able to change it however we feel.

ebrickey: Wouldn't some references to the oldAPI we've yet to revise still call SlavePort::slaveUnbind()?

ebrickey: Actually looking at it closer, it looks like slaveUnbind() is only called in token_port.cc and slaveBind() is never called. In that case I think we could just change the name as you suggested and update the only reference. The description on the function says  "Called by the requestor port to unbind. Should never be called directly." which seems odd since we only found one reference and for only one of the functions?

src/mem/port.cc
```diff
@@ -103,54 +103,55 @@ class DefaultSlavePort : public SlavePort
     AddrRangeList getAddrRanges() const override { return AddrRangeList(); }
 };
 
-DefaultMasterPort defaultMasterPort;
```
powerjg, line 106: Are these variable names used anywhere in gem5?

src/mem/port.hh
```diff
@@ -68,37 +71,37 @@ class SlavePort;
...
-class MasterPort : public Port, public AtomicRequestProtocol,
+class RequestPort: public Port, public AtomicRequestProtocol,
     public TimingRequestProtocol, public FunctionalRequestProtocol
 {
-    friend class SlavePort;
+    friend class ResponsePort;
 
-  private:
-    SlavePort *_slavePort;
+    private:
```
powerjg, line 79: Added spaces here. I know it's weird and IDEs complain, but our style says that private/public/protected are *two* spaces indented.

src/mem/port.hh
```diff
@@ -242,56 +245,63 @@ class MasterPort : public Port, public AtomicRequestProtocol,
     }
 };
 
+class M5_DEPRECATED MasterPort : public RequestPort
+{
+    public:
```
powerjg, line 250: two spaces :)

powerjg (commented):
LGTM! This is perfect so far (assuming things still work ;)). I have a couple of minor comments below.

You may want to get in the habit of running a subset of the regression tests, too. Check out TESTING.md for details (or ask Bobby).

src/mem/port.cc
ebrickey: I searched the gem5 directory for other references and it appears to only be used in port.cc. During our discussion yesterday you renamed a variable and were able to see/rename all other references in the repo. I thought the option was "rename symbol" however when I select this option vscode never successfully finds the references it just has this loading message.
image

src/mem/port.hh
ebrickey: Fixed!

src/mem/port.hh
ebrickey: Fixed! Let me know if that's not quite what you had in mind.

src/mem/port.hh
```diff
@@ -438,71 +448,79 @@ class SlavePort : public Port, public AtomicResponseProtocol,
     }
 };
 
+class M5_DEPRECATED SlavePort : public ResponsePort
+{
+    public:
```
powerjg, line 453: Two spaces instead of four
```suggestion
  public:
```

src/mem/port.hh
```diff
@@ -438,71 +448,79 @@ class SlavePort : public Port, public AtomicResponseProtocol,
...
+        SlavePort(const std::string& name, SimObject* _owner,
```
powerjg, line 454: Four spaces, not eight :)
```suggestion
    SlavePort(const std::string& name, SimObject* _owner,
```

powerjg (commented):
one tiny thing below, but otherwise this is perfect!

src/mem/port.hh
ebrickey: Fixed, thank you for catching this!

src/mem/port.hh
ebrickey: Got it! How do we indent lines where we're past the character limit for a line? For example, like the line below this comment.

powerjg: Very carefully ;). 

I suggest looking within in the file or at other files in the directory and copying that style.

Generally, I would align the next line with the parentheses. The main exception to this would be for `if` statements where you can't tell what is the body and what is the continuation of the condition. Another time this doesn't work is if the parenthesis starts at the end of the line (e.g., >60 chars). 

It's a judgement call... Basically, try to do something that looks the least bad :). The goal is readability. Remember, code is going to be read 10x more than it's written.

PR #181, opened 2020-07-28
SST integration

takekoputa:
This hasn't been finished yet.


ext/sst/Makefile
```diff
@@ -1,9 +1,9 @@
 # These two variables are designed to be modifiable.
-SST_VERSION=SST-trunk
+SST_VERSION=${HOME}/fast/sst/lib/pkgconfig/SST-10.0.0.pc
```
powerjg, line 2: Probably don't want to hardcode this for your setup, generally

takekoputa: yeah, I haven't found a better way so far

PR #183, opened 2020-07-31
Eden/reqresp

eavivi4:
I changed only the parts that used the phrases "MasterPort" or "SlavePort".

### DIFF
--- a/src/dev/arm/SMMUv3.py
+++ b/src/dev/arm/SMMUv3.py
@@ -43,9 +43,9 @@ class SMMUv3SlaveInterface(ClockedObject):
     type = 'SMMUv3SlaveInterface'
     cxx_header = 'dev/arm/smmu_v3_slaveifc.hh'
 
-    slave = SlavePort('Device port')
-    ats_master = MasterPort('ATS master port')
-    ats_slave  = SlavePort('ATS slave port')
+    slave = ResponsePort('Device port')
+    ats_master = RequestPort('ATS master port')
+    ats_slave  = ResponsePort('ATS slave port')
 
     port_width = Param.Unsigned(16, 'Port width in bytes (= 1 beat)')
     wrbuf_slots = Param.Unsigned(16, 'Write buffer size (in beats)')
@@ -74,10 +74,11 @@ class SMMUv3(ClockedObject):
     type = 'SMMUv3'
     cxx_header = 'dev/arm/smmu_v3.hh'
 
-    master = MasterPort('Master port')
-    master_walker = MasterPort(
+    master = RequestPort('Master port')
+    master_walker = RequestPort(
         'Master port for SMMU initiated HWTW requests (optional)')
-    control = SlavePort('Control port for accessing memory-mapped registers')
+    control =
+    ResponsePort('Control port for accessing memory-mapped registers')
     sample_period = Param.Clock('10us', 'Stats sample period')
     reg_map = Param.AddrRange('Address range for control registers')
     system = Param.System(Parent.any, "System this device is part of")
--- a/src/dev/arm/SMMUv3.py
+++ b/src/dev/arm/SMMUv3.py
@@ -77,8 +77,8 @@ class SMMUv3(ClockedObject):
     master = RequestPort('Master port')
     master_walker = RequestPort(
         'Master port for SMMU initiated HWTW requests (optional)')
-    control =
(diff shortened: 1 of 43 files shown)


src/dev/arm/SMMUv3.py
```diff
@@ -74,10 +74,11 @@ class SMMUv3(ClockedObject):
     type = 'SMMUv3'
     cxx_header = 'dev/arm/smmu_v3.hh'
 
-    master = MasterPort('Master port')
-    master_walker = MasterPort(
+    master = RequestPort('Master port')
+    master_walker = RequestPort(
         'Master port for SMMU initiated HWTW requests (optional)')
-    control = SlavePort('Control port for accessing memory-mapped registers')
+    control =
```
powerjg, line 80: This is a bit of a strange way to solve the line length problem. I suggest doing it like the line above and putting `ResponsePort(` on this line and the "help" on the next line.

src/dev/arm/smmu_v3.hh
```diff
@@ -90,7 +90,7 @@ class SMMUv3 : public ClockedObject
     const System &system;
     const MasterID masterId;
 
-    SMMUMasterPort    masterPort;
+    SMMUMasterPort    requestPort;
```
powerjg, line 93: Do you think that "mainPort" would be a better name? Maybe we need to get input from Arm folks to know for sure.

The other ports have more descriptive names... I wonder why this one's name wasn't descriptive...

src/dev/arm/smmu_v3_slaveifc.hh
```diff
@@ -72,7 +72,7 @@ class SMMUv3SlaveInterface : public ClockedObject
     const Cycles microTLBLat;
     const Cycles mainTLBLat;
 
-    SMMUSlavePort *slavePort;
+    SMMUSlavePort *responsePort;
```
powerjg, line 75: Similar to my comment above... I noticed you didn't change the ones below. Maybe this is a case where we should just ask :).

powerjg (commented):
Generally, this looks great!

I have a couple of comments below on the changes to member variable names. However, I don't have a strong opinion or know what the "right" solution is :). I'd love to hear what you think, though.

src/dev/arm/smmu_v3.hh
eavivi4: I don't really know what SMMUMasterPort does, so I don't think I know which one would be better. I think I changed that variable by accident because I thought it's related to MasterPort... Should I leave it as masterPort for now?

powerjg: I'm not sure, tbh. Changing the variable names is definitely harder (and more controversial). That said, we need to do it at some point and get feedback on our names. However, I just have no idea how this is used, so maybe we can skip it for now :)

powerjg (commented):
LGTM!

src/dev/arm/smmu_v3.cc
```diff
@@ -310,7 +310,8 @@ SMMUv3::runProcessTiming(SMMUProcess *proc, PacketPtr pkt)
             DPRINTF(SMMUv3, "[t] master req  addr=%#x size=%#x\n",
                     action.pkt->getAddr(), action.pkt->getSize());
 
-            if (packetsToRetry.empty() && masterPort.sendTimingReq(action.pkt)) {
+            if (packetsToRetry.empty() &&
+            masterPort.sendTimingReq(action.pkt)) {
```
powerjg, line 314: This should be tabbed one or two more times.

powerjg (commented):
One small thing below.

powerjg (commented):
Still lookin' good!

src/dev/arm/smmu_v3.cc
```diff
@@ -310,7 +310,8 @@ SMMUv3::runProcessTiming(SMMUProcess *proc, PacketPtr pkt)
             DPRINTF(SMMUv3, "[t] master req  addr=%#x size=%#x\n",
                     action.pkt->getAddr(), action.pkt->getSize());
 
-            if (packetsToRetry.empty() && masterPort.sendTimingReq(action.pkt)) {
+            if (packetsToRetry.empty() &&
+                    masterPort.sendTimingReq(action.pkt)) {
```
powerjg, line 314: There's nothing wrong with this change :). However, since we are no longer changing the masterPort name here, we should revert this to the original.

PR #184, opened 2020-07-31
Shivani/reu/reqresp

shparekh00:
Changing MasterPort and SlavePort references.
Reverted some changes in src/arch.

### DIFF
--- a/tests/gem5/cpu_tests/ref/FloatMM
+++ b/tests/gem5/cpu_tests/ref/FloatMM
@@ -1,0 +1,6 @@
+gem5 Simulator System.  http://gem5.org
+gem5 is copyrighted software; use the --copyright option for details.
+
+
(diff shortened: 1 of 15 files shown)


tests/gem5/cpu_tests/ref/FloatMM
```diff
@@ -1,6 +0,0 @@
-gem5 Simulator System.  http://gem5.org
```
powerjg, line 1: I don't think you meant to delete this file :D

powerjg (commented):
LGTM except for one thing below (which looks like a mistake :)).

tests/gem5/cpu_tests/ref/FloatMM
shparekh00: Yeah I noticed that and changed it back locally, I'll fix it on my next push!

PR #187, opened 2020-08-04
REU: Update Port Terminology

ebrickey:
This branch is a culmination of all the REU's work to update the port terminology as per Jira: https://gem5.atlassian.net/browse/GEM5-9

### DIFF
--- a/src/arch/arm/ArmTLB.py
+++ b/src/arch/arm/ArmTLB.py
@@ -54,7 +54,7 @@ class ArmTableWalker(ClockedObject):
     # to the Stage2MMU, and shared by the two table walkers, but we
     # access it through the ITB and DTB walked objects in the CPU for
     # symmetry with the other ISAs.
-    port = MasterPort("Port used by the two table walkers")
+    port = RequestPort("Port used by the two table walkers")
 
     sys = Param.System(Parent.any, "system object parameter")
 
--- a/src/arch/generic/BaseTLB.py
+++ b/src/arch/generic/BaseTLB.py
@@ -34,4 +34,4 @@ class BaseTLB(SimObject):
     cxx_header = "arch/generic/tlb.hh"
     # Ports to connect with other TLB levels
     slave  = VectorSlavePort("Port closer to the CPU side")
-    master = MasterPort("Port closer to memory side")
+    master = RequestPort("Port closer to memory side")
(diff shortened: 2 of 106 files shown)


src/dev/arm/smmu_v3.cc
```diff
@@ -311,7 +311,7 @@ SMMUv3::runProcessTiming(SMMUProcess *proc, PacketPtr pkt)
                     action.pkt->getAddr(), action.pkt->getSize());
 
             if (packetsToRetry.empty() &&
-            masterPort.sendTimingReq(action.pkt)) {
+                    masterPort.sendTimingReq(action.pkt)) {
```
powerjg, line 314: This needs to be reverted :)

src/arch/arm/fastmodel/FastModel.py
```diff
@@ -75,21 +75,21 @@ def __init__(self, width, desc):
         super(VectorAmbaInitiatorSocket, self).__init__(
                 my_role, desc, is_source=True)
 
-class ScMasterPort(Port):
```
powerjg, line 78: Let's back out this change. This is changing something quite different than everything else in this commit.

powerjg (commented):
A few changes before we push :). It's better to do it now then wait for someone else to say the same things....

In  https://github.com/darchr/gem5/pull/187/commits/5bcdabf5628b31c4b71ca6448f6ac2372a9dad6b
- the Learning gem5 changes should be in the Learning gem5 commit. It should be in https://github.com/darchr/gem5/pull/187/commits/9c3d2c3be132cdd343db969e35cae5df4484dd77
- It should be titled "mem,sim: ..."

For https://github.com/darchr/gem5/pull/187/commits/ef3333eb2e297624166850e0e5f11c499c7f28a3

Let's update the commit message to
"mem: deprecate SlavePort and MasterPort classes

After this change, if you use these classes or inherit from these classes, the compiler will now give you a warning that these names are deprecated. Instead, you should use ResponsePort and RequestPort, respectively.

This patch simply deprecates these names. The following patches will convert all of the code in gem5 to use these new names. The first step is converting the class names and the uses of these classes, then we will update the variable names to be more precise as well."

src/arch/arm/fastmodel/FastModel.py
shparekh00: So should I keep it as a separate commit?
Or undo the changes?

powerjg: Sorry for the slow reply here :). Probably back out... this is one that we might have to do something special for

PR #189, opened 2020-08-06
Renaming Classes, Changing MasterID and MasterInfo

eavivi4:
Changing classes that contain Master/SlavePort and changing MasterID and MasterInfo into UniqueID and RequestorInfo respectively.


powerjg:
@shparekh00 and @LakinS I've tagged you as reviewers. Could at least one of you review these changes? Thanks!

src/dev/arm/smmu_v3_ports.cc
```diff
@@ -138,7 +138,7 @@ SMMUControlPort::getAddrRanges() const
 
 SMMUATSMasterPort::SMMUATSMasterPort(const std::string &_name,
```
shparekh00: Do we need to change this class too?

src/dev/x86/intdev.hh
```diff
@@ -94,7 +94,7 @@ buildIntPacket(Addr addr, T payload)
 }
 
 template <class Device>
-class IntMasterPort : public QueuedMasterPort
+class IntMasterPort : public QueuedRequestPort
```
shparekh00: Should this be IntRequestPort?

src/mem/cache/base.cc
```diff
@@ -66,7 +66,7 @@ using namespace std;
 BaseCache::CacheSlavePort::CacheSlavePort(const std::string &_name,
```
shparekh00: CacheSlavePort -> CacheResponsePort

src/mem/cache/base.hh
```diff
@@ -119,7 +119,7 @@ class BaseCache : public ClockedObject
      * and the sendDeferredPacket of the timing port is modified to
      * consider both the transmit list and the requests from the MSHR.
      */
-    class CacheMasterPort : public QueuedMasterPort
+    class CacheMasterPort : public QueuedRequestPort
```
shparekh00: CacheRequestPort

src/mem/coherent_xbar.hh
```diff
@@ -82,7 +82,7 @@ class CoherentXBar : public BaseXBar
      * be instantiated for each of the master ports connecting to the
      * crossbar.
      */
-    class CoherentXBarSlavePort : public QueuedSlavePort
+    class CoherentXBarSlavePort : public QueuedResponsePort
```
shparekh00: I think CoherentXBarSlavePort needs to be changed

PR #192, opened 2020-08-11
C++ Variable Updates

shparekh00:
Updating c++ variables and parameters that use slave and master.


src/mem/request.hh
```diff
@@ -302,7 +302,7 @@ class Request
     /** The requestor ID which is unique in the system for all ports
      * that are capable of issuing a transaction
      */
-    MasterID _masterId = invldMasterId;
```
powerjg, line 305: This class should be in its own commit

src/sim/system.cc
```diff
@@ -420,10 +420,10 @@ System::isMemAddr(Addr addr) const
 }
 
 void
-System::addDeviceMemory(MasterID masterId, AbstractMemory *deviceMemory)
+System::addDeviceMemory(MasterID _id, AbstractMemory *deviceMemory)
```
powerjg, line 423: Let's use the same name here as in the header file (uniqueId?)

powerjg (commented):
The changes in request.hh and mem_master.hh should probably be in their own commit(s?). Otherwise, there's one small thing, but it looks good to me generally

PR #193, opened 2020-08-12
Documenting Port API's

takekoputa:
This is not finished yet.

Documented:
- Port / PortProxy / TranslatingPortProxy
- MasterPort / SlavePort
- ExternalMaster / ExternalSlave
- <s>RubyPortProxy</s>

A few notes:
- I don't know users need access to Ruby beyond the RubyPortProxy. I don't know whether RubyPort itself should be a part of the API's. If this is the case then should we also documenting PortProxy etc.

### DIFF
--- a/src/mem/external_master.hh
+++ b/src/mem/external_master.hh
@@ -71,7 +71,7 @@ class ExternalMaster : public SimObject
 
       public:
         /**
-         * @ingroup api_port
+         * @ingroup api_external
          * @{
          */
         ExternalPort(const std::string &name_,
@@ -80,7 +80,7 @@ class ExternalMaster : public SimObject
         { }
 
         ~ExternalPort() { }
-        /** @} */ // end of api_port
+        /** @} */ // end of api_external
 
         /** Any or all of recv... can be overloaded to provide the port's
          *  functionality */
@@ -97,7 +97,7 @@ class ExternalMaster : public SimObject
          * Create or find an external port which can be bound. Returns
          * NULL on failure
          *
-         * @ingroup api_port
+         * @ingroup api_external
          */
         virtual ExternalPort *getExternalPort(
             const std::string &name, ExternalMaster &owner,
@@ -125,14 +125,14 @@ class ExternalMaster : public SimObject
 
   public:
     /**
-      * @ingroup api_port
+      * @ingroup api_external
       */
     ExternalMaster(ExternalMasterParams *params);
 
     /**
      * Port interface.  Responds only to port "port"
      *
-     * @ingroup api_port
+     * @ingroup api_external
      */
     Port &getPort(const std::string &if_name,
                   PortID idx=InvalidPortID) override;
@@ -141,7 +141,7 @@ class ExternalMaster : public SimObject
      * Register a handler which can provide ports with port_type ==
      * handler_name
      *
-     * @ingroup api_port
+     * @ingroup api_external
      */
     static void registerHandler(const std::string &handler_name,
         Handler *handler);
--- a/src/mem/external_master.hh
+++ b/src/mem/external_master.hh
@@ -70,12 +70,17 @@ class ExternalMaster : public SimObject
         ExternalMaster &owner;
 
       public:
+        /**
+         * @ingroup api_port
+         * @{
+         */
         ExternalPort(const std::string &name_,
             ExternalMaster &owner_) :
             MasterPort(name_, &owner_), owner(owner_)
         { }
 
         ~ExternalPort() { }
+        /** @} */ // end of api_port
 
         /** Any or all of recv... can be overloaded to provide the port's
          *  functionality */
@@ -88,8 +93,12 @@ class ExternalMaster : public SimObject
     class Handler
     {
       public:
-        /** Create or find an external port which can be bound.  Returns
-         *  NULL on failure */
+        /**
+         * Create or find an external port which can be bound. Returns
+         * NULL on failure
+         *
+         * @ingroup api_port
+         */
         virtual ExternalPort *getExternalPort(
             const std::string &name, ExternalMaster &owner,
             const std::string &port_data) = 0;
@@ -115,19 +124,36 @@ class ExternalMaster : public SimObject
     static std::map<std::string, Handler *> portHandlers;
 
   public:
+    /**
+      * @ingroup api_port
+      */
     ExternalMaster(ExternalMasterParams *params);
 
-    /** Port interface.  Responds only to port "port" */
+    /**
+     * Port interface.  Responds only to port "port"
+     *
+     * @ingroup api_port
+     */
     Port &getPort(const std::string &if_name,
                   PortID idx=InvalidPortID) override;
 
-    /** Register a handler which can provide ports with port_type ==
-     *  handler_name */
+    /**
+     * Register a handler which can provide ports with port_type ==
+     * handler_name
+     *
+     * @ingroup api_port
+     */
     static void registerHandler(const std::string &handler_name,
         Handler *handler);
 
+    /**
+      * @ingroup api_port
+      */
     void init() override;
 
+    /**
+      * @ingroup api_port
+      */
     const MasterID masterId;
 };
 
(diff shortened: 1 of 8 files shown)


src/mem/external_master.hh
```diff
@@ -115,19 +124,36 @@ class ExternalMaster : public SimObject
...
+     *
+     * @ingroup api_port
+     */
     static void registerHandler(const std::string &handler_name,
         Handler *handler);
 
+    /**
+      * @ingroup api_port
+      */
     void init() override;
```
powerjg, line 152: Not part of this API, it's part of the SimObject API

src/mem/external_master.hh
```diff
@@ -70,12 +70,17 @@ class ExternalMaster : public SimObject
         ExternalMaster &owner;
 
       public:
+        /**
+         * @ingroup api_port
```
powerjg, line 74: External ports are different from the "port" API. Maybe this should be "external_api"?

src/mem/external_master.hh
```diff
@@ -115,19 +124,36 @@ class ExternalMaster : public SimObject
     static std::map<std::string, Handler *> portHandlers;
 
   public:
+    /**
+      * @ingroup api_port
+      */
     ExternalMaster(ExternalMasterParams *params);
```
powerjg, line 130: I think constructors are part of the SimObject API... but I could be wrong. Thoughts @BobbyRBruce?

src/mem/external_master.hh
```diff
@@ -115,19 +124,36 @@ class ExternalMaster : public SimObject
...
 
+    /**
+      * @ingroup api_port
+      */
     void init() override;
 
+    /**
+      * @ingroup api_port
+      */
     const MasterID masterId;
```
powerjg, line 157: Should this be public? Is it used publicly anywhere?

src/mem/external_slave.hh
```diff
@@ -123,17 +132,31 @@ class ExternalSlave : public SimObject
...
+     */
     ExternalSlave(ExternalSlaveParams *params);
 
-    /** Port interface.  Responds only to port "port" */
+    /**
+     * Port interface. Responds only to port "port".
+     *
+     * @ingroup api_port
+     */
     Port &getPort(const std::string &if_name,
```
powerjg, line 145: This should be defined in the sublcass, not here (I think). @BobbyRBruce, do you remember how APIs worked with sublcasses last time?

src/mem/port.hh
```diff
@@ -80,18 +80,27 @@ class MasterPort : public Port, public AtomicRequestProtocol,
     SimObject &owner;
 
   public:
+    /**
+     * @ingroup api_port
+     * @{
+     */
     MasterPort(const std::string& name, SimObject* _owner,
```
powerjg, line 87: FYI, this is going to conflict with https://gem5-review.googlesource.com/c/public/gem5/+/32308. We want to make sure that is merged first. I don't want to define Master/Slave in the API officially.

src/mem/port.hh
```diff
@@ -287,15 +323,20 @@ class SlavePort : public Port, public AtomicResponseProtocol,
...
      * @return a list of ranges responded to
+     *
+     * @ingroup api_port
      */
     virtual AddrRangeList getAddrRanges() const = 0;
 
     /**
      * We let the master port do the work, so these don't do anything.
+     *
+     * @ingroup api_port
```
powerjg, line 334: These should be part of the parent class API, not the API here. Again, let Bobby correct me if I'm wrong here.

src/mem/ruby/system/RubyPortProxy.hh
```diff
@@ -58,17 +58,23 @@ class RubyPortProxy : public RubyPort
...
+     *
+     * @ingroup api_port
      */
     virtual ~RubyPortProxy();
 
     /**
      * Initialise a RubyPortProxy by doing nothing and avoid
      * involving the super class.
+     *
+     * @ingroup api_port
```
powerjg, line 77: part of SimObj API

src/mem/translating_port_proxy.hh
```diff
@@ -72,18 +72,31 @@ class TranslatingPortProxy : public PortProxy
...
 
-    /** Version of tryReadblob that translates virt->phys and deals
-      * with page boundries. */
+    /**
+     * Version of tryReadblob that translates virt->phys and deals
+     * with page boundries.
+     *
+     * @ingroup api_port
+     */
     bool tryReadBlob(Addr addr, void *p, int size) const override;
```
powerjg, line 86: In the parent's API?

takekoputa:
> Generally, I think that RubyPortProxy should not be part of the API, but PortProxy should be.

I saw it being used in learning gem5 part 3: https://gem5.googlesource.com/public/gem5/+/refs/heads/master/configs/learning_gem5/part3/test_caches.py#96

Is there any reason why RubyPortProxy shouldn't be part of the API?

powerjg:
Hmm... good point. Let me mull it over a bit

src/mem/external_master.hh
takekoputa: I find this quite confusing. Even though it's a part of SimObject API, but it has its own signature, i.e. ExternalMasterParams being the parameter.

src/mem/external_master.hh
takekoputa: I agree.

src/mem/port.hh
takekoputa: Sure, it's gonna be a huge merge conflict. I'd wait until the terminology change to be done before pushing this to gerrit.

src/mem/external_master.hh
powerjg: Yeah... So, the constructors should actually never be used outside of the *Params::create() methods. In some sense, they shouldn't be "public"... In fact, now that I write this, they shouldn't be public! The *Params class should be a friend.

Let's not do make that change now... However, let's not include the constructor as part of the API.

powerjg:
No, RubyPortProxy is just a specific SimObject. There's no need to mark it as an "important" API (same with RubyPort)

src/mem/external_master.hh
takekoputa: Done

src/mem/external_master.hh
takekoputa: Removed masterId from the API.

takekoputa: anyway, a function in util/tlm/src/sc_master_port.cc  uses masterId to make a request

src/mem/external_slave.hh
takekoputa: done

src/mem/external_master.hh
takekoputa: done

src/mem/port.hh
takekoputa: I'm trying to be consistent here. I assume that bind() and unbind() of Master are also not in the Port API.

src/mem/translating_port_proxy.hh
takekoputa: done

src/mem/ruby/system/RubyPortProxy.hh
takekoputa: done

takekoputa:
So should RubyPort a part of the port API? It isn't derived from Port, but it contains 2 subclasses: MemMasterPort and MemSlavePort that are inherited from QueuedMasterPort/QueuedSlavePort.

powerjg:
No, RubyPort shouldn't be part of the PortAPI... it's quite different

src/mem/external_master.hh
```diff
@@ -117,12 +126,15 @@ class ExternalMaster : public SimObject
...
     ExternalMaster(ExternalMasterParams *params);
 
-    /** Port interface.  Responds only to port "port" */
     Port &getPort(const std::string &if_name,
                   PortID idx=InvalidPortID) override;
 
-    /** Register a handler which can provide ports with port_type ==
-     *  handler_name */
+    /**
+     * Register a handler which can provide ports with port_type ==
```
powerjg, line 133: Can you provide any other context here? Do you know how this handler might be used?

PR #194, opened 2020-08-17
arch-arm: convert tlb to new style stats

powerjg:
Change-Id: I2a3f138b53496be6361a1a2b81fa471a56a4dc10

### DIFF
--- a/src/arch/arm/tlb.cc
+++ b/src/arch/arm/tlb.cc
@@ -76,7 +76,7 @@ TLB::TLB(const ArmTLBParams *p)
     : BaseTLB(p), table(new TlbEntry[p->size]), size(p->size),
       isStage2(p->is_stage2), stage2Req(false), stage2DescReq(false), _attr(0),
       directToStage2(false), tableWalker(p->walker), stage2Tlb(NULL),
-      stage2Mmu(NULL), test(nullptr), rangeMRU(1),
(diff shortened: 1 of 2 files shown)


src/arch/arm/tlb.cc
```diff
@@ -414,121 +414,34 @@ TLB::takeOverFrom(BaseTLB *_otlb)
...
+    ADD_STAT(alignFaults, "Number of TLB faults due to alignment"
+        " restrictions"),
+    ADD_STAT(prefetchFaults, "Number of TLB faults due to prefetch"),
+    ADD_STAT(domainFaults, "Number of TLB faults due to domain restrictions"),
+    ADD_STAT(permsFaults, "Number of TLB faults due to permissions"
+        " restrictions"),
+    ADD_STAT(readAccesses, "DTB read accesses", readHits + readMisses),
+    ADD_STAT(writeAccesses, "DTB write accesses", writeHits + writeMisses),
+    ADD_STAT(instAccesses, "ITB inst accesses", instHits + instMisses),
+    ADD_STAT(hits, "DTB hits", readHits + writeHits + instHits),
```
powerjg, line 440: For the description, I think "Total TLB (inst and data) hits" would be better. Same for the next 2.

ebrickey: Gotcha it's been updated!

PR #201, opened 2020-08-25
Function update

shparekh00:
This branch will contain terminology updates specifically to function names that use master/slave.


src/mem/request.hh
```diff
@@ -654,7 +654,7 @@ class Request
 
     /** Accesssor for the requestor id. */
     MasterID
-    masterId() const
+    uniqueId() const
```
powerjg, line 657: I think this should be "requestorId"

src/mem/packet.hh
```diff
@@ -706,7 +706,7 @@ class Packet : public Printable
     inline void qosValue(const uint8_t qos_value)
     { _qosValue = qos_value; }
 
-    inline MasterID masterId() const { return req->masterId(); }
+    inline MasterID uniqueId() const { return req->uniqueId(); }
```
powerjg, line 709: This too should be "requestorId"

src/mem/request.hh
shparekh00: If I name the function here requestorId, it'll have to be called that everywhere else, so I thought uniqueId made more sense as the MasterID class is now UniqueID.

Otherwise it would be an easy fix if you prefer requestorId()

PR #203, opened 2020-08-26
Remove mem_slave_port from Sequencer

LakinS:
I just removed the mem_slave_port param from the Sequencer file and it's .cc file. 

Change-Id: I653119d435c5c01d1a3ad7222d2c838845d59922


powerjg (commented):
This is OK, but it conflicts with what's on gerrit right now.

Go ahead and push to gerrit, but wait until after Shivani's large changeset is committed :).

One more little thing: Make a point in the commit message that this port is unused, which is why you're removing it.

PR #204, opened 2020-08-26
Allow pre-commit hook to parse unicode/nonASCII characters

LakinS:
Set the default encoding in python to utf-8

Change-Id: I76ff6e573b9718d3023177bbd50b37a6ee3bc514


powerjg (commented):
Perfect! Push to gerrit :D

PR #207, opened 2020-10-29
Updates in Ticked stats [wip]

aakahlow:
This change will remove the TickedObject and move Ticked to cpu/minor/ as it is only used by MinorCPU and update the stats style of Ticked class.


ayaz91:
This still has some problems. Basically, even before the third commit in the above change is applied (the updated stats one), I get some pybind errors while compiling gem5:

"ext/pybind11/include/pybind11/pybind11.h:1057:5: error: static assertion failed: Unknown/invalid class_ template parameters provided                                   
     static_assert(detail::all_of<is_valid_class_option<options>...>::value,
"

Will appreciate any input.

src/sim/ticked.cc
```diff
@@ -35,9 +35,9 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include "sim/ticked_object.hh"
+#include "sim/ticked.hh"
 
-#include "params/TickedObject.hh"
+#include "params/Ticked.hh"
```
powerjg: This doesn't seem right to me. I think this is still TickedObject since that's the name of the SimObject in Ticked.py

src/sim/Ticked.py
```diff
@@ -36,6 +36,6 @@
 from m5.objects.ClockedObject import ClockedObject
 
 class TickedObject(ClockedObject):
```
powerjg: Why do we still have a "TickedObject" in python?

src/cpu/minor/Ticked.py
```diff
@@ -35,7 +35,7 @@
 
 from m5.objects.ClockedObject import ClockedObject
 
-class TickedObject(ClockedObject):
+class Ticked(ClockedObject):
```
powerjg, line 38: I think this should be in the previous commit

powerjg (commented):
Looks great! How did you test this?

powerjg:
How did you test? I'll try to figure out your error on my machine :)

ayaz91:
This is not even compiling yet. I tried to compile for ARM (as it compiles for MinorCPU by default)

src/cpu/minor/ticked.cc
```diff
@@ -47,39 +47,29 @@ Ticked::Ticked(ClockedObject &object_,
...
 }
 
-void
-Ticked::regStats()
+Ticked::TickedStats::TickedStats(Stats::Group *parent)
+    : Stats::Group(parent, "Ticked"),
+      ADD_STAT(tickCycles,
+                    "Number of cycles that the object actually ticked."),
+      ADD_STAT(idleCycles, "Total number of cycles that the object"
+                    "has spent stopped.",numCycles - tickCycles)
```
powerjg, line 70: This is causing a compiler error. You can't use `numCycles` which is not a member of `TickedStats`.

```suggestion
Ticked::TickedStats::TickedStats(Ticked *parent)
    : Stats::Group(parent, "Ticked"),
      ADD_STAT(tickCycles,
                    "Number of cycles that the object actually ticked."),
      ADD_STAT(idleCycles, "Total number of cycles that the object"
                    "has spent stopped.", parent->numCycles - tickCycles)
```

powerjg:
Ok. With those two changes, I'm now seeing the same error you described.

powerjg:
Ok, so I understand the problem now.

You have `class Ticked(ClockedObject)` in the SimObject description file, but in C++ you have `class Ticked : public Serializable, public Stats::Group`.

You have to match the hierarchy in Python and C++. In Python `Ticked` inherits from `ClockedObject`, but in C++ it doesn't. Which one of these inheritances is correct?

ayaz91:
> You have to match the hierarchy in Python and C++. In Python Ticked inherits from ClockedObject, but in C++ it doesn't. Which one of these inheritances is correct?

Sorry, my bad. Ticked inherits from Serializable and not from ClockedObject. TickedObject was inheriting from Ticked and ClockedObject. When I moved TickedObject to Ticked, I did not take care of this

powerjg:
Does it work now? :)

src/cpu/minor/Ticked.py
```diff
@@ -35,7 +35,9 @@
 
 from m5.objects.ClockedObject import ClockedObject
 
-class TickedObject(ClockedObject):
-    type = 'TickedObject'
+class Ticked():
```
powerjg, line 38: This definitely won't work. It's not a SimObject.

Does this need to be a Python object?

ayaz91: I don't think that it needs to be  a Python object. Do we need to create  a Python object if it has no associated parameters or it is never going to be accessed by the user through Python interface?

ayaz91:
> Does it work now? :)
This compiles now but ticked and idle cycles stats do not show in the stats file when a simulation is run

src/cpu/minor/Ticked.py
powerjg: Yeah, we can probably drop the python completely.

powerjg:
We're getting somewhere! I can look at this a bit tomorrow :).

src/cpu/minor/ticked.hh
```diff
@@ -39,23 +39,20 @@
...
 
-class TickedObjectParams;
-
 /** Ticked attaches gem5's event queue/scheduler to evaluate
  *  calls and provides a start/stop interface to ticking.
  *
  *  Ticked is not a ClockedObject but can be attached to one by
  *  inheritance and by calling regStats, serialize/unserialize */
-class Ticked : public Serializable
+class Ticked : public Serializable, public Stats::Group
```
powerjg, line 55: Does this need to be a Stats::Group? Would it make more sense to pass the CPU in as the parent to the struct?

ayaz91: Yeah, it does not need to inherit from Stats::Group. I have pushed a change to do that. Ticked's implementation already seems tied to minor CPU, so I don't think this will be use in other ways

PR #208, opened 2020-11-04
Adding sources for MemScheduler and LLM interface.

marjanfariborz:


powerjg:
Much easier to review! I'll take a look later today :)

powerjg:
What is this a branch from? It shouldn't be master. Should it be stable (v20.1)? Develop?

Please rebase so I can review more easily. For instance, right now under "files" it has DRAMInterface.py as a change in this PR. I don't think that is correct.

src/mem/mem_scheduler.cc
```diff
@@ -0,0 +1,435 @@
...
+bool
+MemScheduler::handleRequest(CPUSidePort *port, PacketPtr pkt)
+{
+    panic_if(!(pkt->isRead() || pkt->isWrite()),
+             "Should only see read and writes at memory controller\n");
+
+    uint32_t requestorId = pkt->req->requestorId();
+
+    std::unordered_map<RequestorID, CPUSidePort*>::const_iterator ret = routingTable.find(requestorId);
+    if (ret == routingTable.end())
```
powerjg, line 217: Please use curly braces around all if statements. Here is a concrete example of a bug caused by not having curly braces: https://gem5-review.googlesource.com/c/public/gem5/+/29027/8/src/mem/dram_ctrl.cc#813. Compare base line 694 with new line 813.

src/mem/mem_scheduler.cc
```diff
@@ -0,0 +1,435 @@
...
+    std::unordered_map<RequestorID, bool>::const_iterator rit = readBlocked.find(requestorId);
+    if (rit == readBlocked.end())
+        readBlocked[requestorId] = false;
+
+    std::unordered_map<RequestorID, bool>::const_iterator wit = writeBlocked.find(requestorId);
+    if (wit == writeBlocked.end())
+        writeBlocked[requestorId] = false;
+
+    if (pkt->isRead() && readBlocked[requestorId])
+        return false;
```
powerjg, line 229: Could this be a bug? Is it a problem to add something to the routing tables and update the blocked tables *then* return that the packet can't be handled.

Usually, you want to check to see if it can be serviced first, and then start updating things.

I could be wrong. This is just another smell :)

src/mem/mem_scheduler.cc
```diff
@@ -0,0 +1,435 @@
...
+        DPRINTF(MemScheduler, "Scheduling nextReqEvent in handleRequest\n");
+        schedule(nextReqEvent, curTick());
+    }
+    return true;
+}
+
+MemScheduler::MemSidePort*
+MemScheduler::findMemoryPort(PacketPtr pkt){
+    const Addr base_addr = pkt->getAddr();
+    for (auto memPort : memPorts)
```
powerjg, line 264: There is an `AddrRangeMap` for this :D.

src/mem/mem_scheduler.cc
```diff
@@ -0,0 +1,435 @@
...
+            if (addr_range.contains(base_addr)){
+                return memPort;
+            }
+    return NULL;
+}
+
+void
+MemScheduler::processNextReqEvent(){
+    std::unordered_map<RequestorID, std::queue<PacketPtr> >::iterator \
+                                        initialEntry = currentReadEntry;
```
powerjg, line 275: `auto` is your friend!
```suggestion
    auto initialEntry = currentReadEntry;
```

src/mem/mem_scheduler.cc
```diff
@@ -0,0 +1,435 @@
...
+    CPUSidePort *cpuPort;
+    MemSidePort *port;
+    PacketPtr pkt;
+    RequestorID requestorId;
+    bool sendRetry = false;
+    // DPRINTF(MemScheduler, "processNextReqEvent: just entered\n");
+    while (true){
```
powerjg, line 282: I'd be surprised if using a while/break was the best implementation. Can you describe what you're trying to do here? I may be able to suggest a better implementation.

src/mem/mem_scheduler.hh
```diff
@@ -0,0 +1,290 @@
+/*
+ * Copyright (c) 2017 Jason Lowe-Power
```
powerjg, line 2: UC Regents

src/mem/mem_scheduler.hh
```diff
@@ -0,0 +1,290 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __LEARNING_GEM5_PART2_SIMPLE_MEMOBJ_HH__
+#define __LEARNING_GEM5_PART2_SIMPLE_MEMOBJ_HH__
```
powerjg, line 30:
```suggestion
#ifndef __MEM_MEM_SCHEDULER_HH__
#define __MEM_MEM_SCHEDULER_HH__
```

powerjg:
No!!! I'm commenting on this one :S

src/mem/mem_scheduler.hh
```diff
@@ -0,0 +1,290 @@
...
+#include <unordered_map>
+#include <queue>
+
+#include "mem/port.hh"
+#include "params/MemScheduler.hh"
+#include "sim/sim_object.hh"
+#include "sim/clocked_object.hh"
+
+/**
+ * A very simple memory object. Current implementation doesn't even cache
```
powerjg, line 41: This comment is wrong. Please update with current info.

src/mem/mem_scheduler.hh
```diff
@@ -0,0 +1,290 @@
...
+     * @return the address ranges this memobj is responsible for
+     */
+    AddrRangeList getAddrRanges() const;
+
+    /**
+     * Tell the CPU side to ask for our memory ranges.
+     */
+    void sendRangeChange();
+
+    void wakeUp();
```
powerjg, line 235: What does this function do?

src/mem/mem_scheduler.hh
```diff
@@ -0,0 +1,290 @@
...
+    const uint32_t readBufferSize;
+    const uint32_t writeBufferSize;
+    const uint32_t respBufferSize;
+    const uint32_t numberPorts;
+    const uint32_t numberQueues;
+    bool respBlocked;
+
+    // std::queue<PacketPtr> *readQueue;
+    // std::queue<PacketPtr> *writeQueue;
+    std::unordered_map<RequestorID, std::queue<PacketPtr> >::iterator currentReadEntry;
```
powerjg, line 260: Why needed?

powerjg (commented):
Some comments :). It probably makes sense to sit down together and go through all of this.

marjanfariborz:
DRAMInterface.py has been changed. I needed to add LLM to the DRAMInterface.py.

powerjg:
Sure, but you didn't add over 1000 lines, right? This looks like it should have been a PR based on a different branch (e.g., stable). What is the base commit for this change?

Click the "files change" tab at the top of this PR. Look to see if it has the changes as you expect them to look. If not, try to fix it. I can help figure out exactly what to do if needed :).

src/mem/mem_scheduler.hh
marjanfariborz: When all of the memory side ports are blocked and all the queues in the memory scheduler are full then the nextReqEvent won't be scheduled. We need a wakeup function to schedule the nextReqEvent when memory side ports are no longer blocked.

src/mem/mem_scheduler.hh
marjanfariborz: We use this as an arbiter over all the read queues in the memory scheduler

src/mem/mem_scheduler.hh
```diff
@@ -0,0 +1,291 @@
...
+#include "sim/clocked_object.hh"
+
+/**
+ * A very simple memory object. Current implementation doesn't even cache
+ * anything it just forwards requests and responses.
+ * This memobj is fully blocking (not non-blocking). Only a single request can
+ * be outstanding at a time.
+ */
+// typedef std::deque<MemPacket*> MemPacketQueue;
+class MemScheduler : public SimObject
```
powerjg, line 47: Should this be a ClockedObject?

src/mem/mem_scheduler.cc
```diff
@@ -0,0 +1,499 @@
...
+    owner->sendRangeChange();
+}
+
+bool
+MemScheduler::handleRequest(CPUSidePort *port, PacketPtr pkt)
+{
+    panic_if(!(pkt->isRead() || pkt->isWrite()),
+             "Should only see read and writes at memory controller\n");
+
+    uint32_t requestorId = pkt->req->requestorId();
```
powerjg, line 214:
```suggestion
    RequestorID requestorId = pkt->req->requestorId();
```

src/mem/mem_scheduler.cc
```diff
@@ -0,0 +1,499 @@
...
+
+    std::unordered_map<RequestorID, CPUSidePort*>::const_iterator ret = routingTable.find(requestorId);
+    if (ret == routingTable.end())
+        routingTable[requestorId] = port;
+
+    std::unordered_map<RequestorID, bool>::const_iterator rit = readBlocked.find(requestorId);
```
powerjg, line 220: Initializing these maps should probably be done in `::startup()`. I believe that you can query the system object (you can add this as a parameter with `parent.Any` as the default proxy) to get the number of requestors in the system.

PR #212, opened 2020-12-16
Python stats

powerjg:
Maybe this will work...


src/python/m5/simstats/jsonserializable.py
```diff
@@ -0,0 +1,54 @@
...
+class JsonSerializable:
+
+    # Must be in accordance to the JSON Schema
+    def to_json_dict(self) -> dict:
+        model_dct = {}
+        for key, value in self.__dict__.items():
+            model_dct[key] = process_value(value)
+        return model_dct
+
+def process_value(value):
```
powerjg, line 41: Should this me a member of JsonSerializable? Maybe as a static class method?

Alternatively, if it's supposed to be a "private" function, you can lead with an underscore for the name.

src/python/m5/simstats/jsonserializable.py
```diff
@@ -0,0 +1,54 @@
...
+        return value
+    elif isinstance(value, datetime):
+        return value.replace(microsecond=0).isoformat()
+    elif isinstance(value, list):
+        to_return = []
+
+        for v in value:
+            to_return.append(process_value(v))
+
+        return to_return
```
powerjg, line 54: I'd do
```suggestion
        return [process_value(v) for v in value]
```

src/python/m5/simstats/statshandler.py
```diff
@@ -0,0 +1,94 @@
...
+    return Model(simobj_type, time_conversion, **stats_dict)
+
+def _get_statistic(statistic):
+    # Right now I'll just focus on scalars. # TODO
+    if isinstance(statistic, _m5.stats.ScalarInfo):
+        value = statistic.value()
+        stat_type = "Scalar"
+        unit = None # TODO
+        description = statistic.desc
+        datatype = None # TODO
```
powerjg, line 64: I think this is something like uint64. Can double check in the C++, though :)

powerjg (commented):
A couple of small things :)

src/python/m5/simstats/jsonserializable.py
BobbyRBruce: Done.

src/python/m5/simstats/jsonserializable.py
BobbyRBruce: Done

src/python/m5/simstats/statshandler.py
BobbyRBruce: Yeah, this should be easy to figure out, i just haven't double checked.

PR #217, opened 2021-03-10
arch-riscv: PMP Implementation

aakahlow:
Still in progress.


src/arch/riscv/RiscvTLB.py
```diff
@@ -44,6 +44,7 @@ class RiscvPagetableWalker(ClockedObject):
             "Number of outstanding walks that can be squashed per cycle")
     # Grab the pma_checker from the MMU
     pma_checker = Param.PMAChecker(Parent.any, "PMA Chekcer")
+    pmp = Param.PMP(Parent.any, "PMP")
```
powerjg, line 47: I think this makes sense for now, but I'd like to see a change after this one which factors out these permission checks so we don't have do duplicate the code in both the TLB and the PTW.

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,60 @@ ISA::setMiscReg(int misc_reg, RegVal val)
         warn("Ignoring write to %s.\n", CSRData.at(misc_reg).name);
     } else {
         switch (misc_reg) {
+
+          case MISCREG_PMPCFG0:
```
powerjg, line 338: Can you explain this fallthrough. It's probably obvious from the spec, but why do we handle CFG0 and CFG2 the same (what about CFG1??). A brief statement and a pointer to the spec is probably enough.

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,60 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+          case MISCREG_PMPCFG2:
+            {
```
powerjg, line 340: Not sure if the spacing is right on this line.

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,60 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+
+                uint8_t cfg_val = (val >> 8 * i)  & 0xff;
+                auto mmu = dynamic_cast<RiscvISA::MMU *>(tc->getMMUPtr());
+
+                auto pp = dynamic_cast<PMP *>(mmu->getPMP());
+
+                pp->pmp_update_cfg(i+(4*(misc_reg-MISCREG_PMPCFG0)),
+                                  cfg_val);
+                }
+                setMiscRegNoEffect(misc_reg, val);
```
powerjg, line 365: Odd whitespace. `{` go on the same line and the for loop

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,60 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+                std::cout << "PMP cfg update " << std::endl;
+
+                int i;
+                for (i=0; i < sizeof(val); i++)
+                {
+
+                uint8_t cfg_val = (val >> 8 * i)  & 0xff;
+                auto mmu = dynamic_cast<RiscvISA::MMU *>(tc->getMMUPtr());
+
+                auto pp = dynamic_cast<PMP *>(mmu->getPMP());
```
powerjg, line 360: Shouldn't need the cast here since the mmu is a RiscvISA::MMU

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,60 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+
+                pp->pmp_update_cfg(i+(4*(misc_reg-MISCREG_PMPCFG0)),
+                                  cfg_val);
+                }
+                setMiscRegNoEffect(misc_reg, val);
+            }
+            break;
+
+          case MISCREG_PMPADDR00 ... MISCREG_PMPADDR15:
```
powerjg, line 369: :open_mouth: What is this syntax?!?!

Very cool.. but I've never seen it before!

I think this is GCC-specific. The spec doesn't mention this: https://en.cppreference.com/w/cpp/language/switch

GCC does talk about it, though: https://gcc.gnu.org/onlinedocs/gcc/Case-Ranges.html

I think we need to not use this syntax since it's not part of the C++ spec. Feel free to disagree, though :)

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,60 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+                // In qemu, the rule is updated whenever
+                // pmpaddr is written
+                // (in other words, pmpcfg/addr are converted)
+                // into a rule
+
+                assert(readMiscRegNoEffect(MISCREG_PRV) == PRV_M);
+
+                auto mmu = dynamic_cast<RiscvISA::MMU *>(tc->getMMUPtr());
+
+                auto pp = dynamic_cast<PMP *>(mmu->getPMP());
```
powerjg, line 383: cast shouldn't be needed

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,29 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
             Addr paddr = walker->tlb->translateWithTLB(vaddr, satp.asid, mode);
             req->setPaddr(paddr);
             walker->pma->check(req);
+
+            if (pmode != PrivilegeMode::PRV_M){
+                bool pass = walker->pmp->pmp_check(req, mode);
```
powerjg, line 495: what does "pass" mean?

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,29 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
...
+                // what else to do here?
+                // raise some kind of an exception
+
+                if (!pass)
+                {
```
powerjg, line 500: On previous line

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,29 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
...
+                    ExceptionCode code;
+                    if (mode == TLB::Read)
```
powerjg, line 502: I strongly prefer having `{ }` even on single line `if` statements

powerjg, line 502: Actually, I wonder if this code should be factored out into a function (e.g., `createMemException(TLB::Mode)`)

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,62 @@
...
+#define __ARCH_RISCV_PMP_HH__
+
+#include "arch/generic/tlb.hh"
+#include "base/addr_range.hh"
+#include "base/types.hh"
+#include "mem/packet.hh"
+#include "params/PMP.hh"
+#include "sim/sim_object.hh"
+
+class PMP : public SimObject
```
powerjg, line 11: Doxygen on this class would be good.

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,62 @@
...
+
+    typedef PMPParams Params;
+
+    const Params &
+    params() const
+    {
+        return dynamic_cast<const Params &>(_params);
+    }
+
+    int Max_Entries;
```
powerjg, line 23: snake case

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,62 @@
...
+
+    // For encoding of address matching
+    // mode of PMP address register
+    // A (3-4) bits of pmpcfg register
+    typedef enum {
+        PMP_OFF,
+        PMP_TOR,
+        PMP_NA4,
+        PMP_NAPOT
+    } pmp_amatch;
```
powerjg, line 35: SnakeCase

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,62 @@
...
+        uint8_t  pmp_cfg;
+    } pmp_entry;
+
+    std::vector<pmp_entry*> pmp_table;
+
+    bool pmp_check(const RequestPtr &req, BaseTLB::Mode mode);
+    static inline uint8_t pmp_get_a_field(uint8_t cfg);
+    void pmp_update_cfg(uint32_t pmp_index, uint8_t this_cfg);
+    void pmp_update_addr(uint32_t pmp_index, Addr this_addr);
+    AddrRange pmp_decode_napot(Addr a);
```
powerjg, line 58: All of these function names should be snake case

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,62 @@
...
+
+    // struct corresponding to a single
+    // PMP entry
+    typedef struct {
+        AddrRange pmp_addr;
+        Addr raw_addr;
+        uint8_t  pmp_cfg;
+    } pmp_entry;
+
+    std::vector<pmp_entry*> pmp_table;
```
powerjg, line 52: snake case (and private?)

src/arch/riscv/tlb.cc
```diff
@@ -366,6 +368,32 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
...
+                    else if (mode == TLB::Write)
+                        code = ExceptionCode::STORE_ACCESS;
+                    else
+                        code = ExceptionCode::INST_ACCESS;
+                    fault = std::make_shared<AddressFault>
+                        (req->getVaddr(), code);
+
+                }
+
+            }
```
powerjg, line 395: So, this code is duplicated in the TLB and in the PTW. This is what I'm talking about when I say it would be good to unify the permission checking. This is OK for now, but should be done in a follow on patch.

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,153 @@
+#include "arch/riscv/pmp.hh"
```
powerjg, line 1: Missing license

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,29 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
             Addr paddr = walker->tlb->translateWithTLB(vaddr, satp.asid, mode);
             req->setPaddr(paddr);
             walker->pma->check(req);
+
```
powerjg, line 493: This file changed permissions. This should (almost) never happen.

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "arch/generic/tlb.hh"
+#include "base/addr_range.hh"
+#include "base/types.hh"
+#include "math.h"
+#include "mem/request.hh"
+#include "params/PMP.hh"
+#include "sim/sim_object.hh"
+
+PMP::PMP(const Params &params) :
+SimObject(params), Max_Entries(params.max_pmp)
```
powerjg, line 12: needs some spaces

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,153 @@
...
+{
+
+ for (int i=0; i<Max_Entries; i++){
```
powerjg, line 15: Spacing is all messed up here...

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,153 @@
...
+
+    pmp_entry* entry = new pmp_entry;
+    pmp_table.push_back(entry);
```
powerjg, line 18: Use `emplace` instead and don't store pointers.

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,153 @@
...
+bool
+PMP::pmp_check(const RequestPtr &req, BaseTLB::Mode mode)
+{
+
+// all pmp entries need to be looked from the lowest to
+// the highest number
+
+    bool addr_match = false;
+
+    int i;
```
powerjg, line 34: this should be inside the for statement

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,153 @@
...
+
+    uint8_t allowed_privs = PMP_READ | PMP_WRITE | PMP_EXEC;
+
+    // i is the index of pmp table which matched
+    allowed_privs &= pmp_table[i]->pmp_cfg;
+
+
+    // How to determine if the request is for an inst cache
+    if ((mode == BaseTLB::Mode::Read) &&
+            ((PMP_READ & allowed_privs) == PMP_READ))
```
powerjg, line 56: `{ }` would be nice

powerjg (commented):
Mostly reviewed, but not 100%. Please make updates and I'll review again once all of my comments have been addressed

src/arch/riscv/isa.cc
ayaz91: I have added a comment in the code. According to specs, the odd-numbered configuration registers are illegal for RV64 and each 64 bit CFG register hold configurations for 8 PMP entries. The only difference in handling of these regs, will be passing the correct pmp index to pmpUpdateCfg function(), which I am trying to do through this expression "i+(4*(misc_reg-MISCREG_PMPCFG0)" where i refers to a PMP entry number.

src/arch/riscv/isa.cc
ayaz91: Fixed this.

src/arch/riscv/isa.cc
ayaz91: Fixed.

src/arch/riscv/isa.cc
ayaz91: tc->getMMUPtr() returns a pointer to the BaseMMU and not RiscvISA::MMU, that's why the cast.

src/arch/riscv/isa.cc
ayaz91: fixed.

src/arch/riscv/pagetable_walker.cc
ayaz91: "pass=true" means that the pmp checks are passed. Added a comment.

src/arch/riscv/pagetable_walker.cc
ayaz91: fixed.

src/arch/riscv/pagetable_walker.cc
ayaz91: fixed.

ayaz91: created a function

src/arch/riscv/pmp.hh
ayaz91: Added a placeholder.

src/arch/riscv/pmp.hh
ayaz91: styling guide seems to suggest these should be camelCase. Is that correct? I have updated the case of all class members

src/arch/riscv/pmp.hh
ayaz91: Same as above.

src/arch/riscv/pmp.hh
ayaz91: Same as above.

src/arch/riscv/pmp.cc
ayaz91: Added.

src/arch/riscv/tlb.cc
ayaz91: Yeah, that still needs to be done.

src/arch/riscv/pmp.cc
ayaz91: fixed.

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,61 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+                std::cout << "PMP cfg update " << std::endl;
+
+                int i;
+                for (i=0; i < sizeof(val); i++) {
+
+                    uint8_t cfgVal = (val >> 8 * i) & 0xff;
+                    auto mmu = dynamic_cast<RiscvISA::MMU *>
+                                (tc->getMMUPtr());
+                    mmu->getPMP()->pmpUpdateCfg
+                                (i+(4*(misc_reg-MISCREG_PMPCFG0)),cfgVal);
```
powerjg, line 366:
```suggestion
                    int pmp_num = i+ 4 * (misc_reg - MISCREG_PMPCFG0)
                    mmu->getPMP()->pmpUpdateCfg(pmp_num, cfgVal);
```
Something like that. I'm not sure I fully understand, but I think it would be clearer if you used a local variable.

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,61 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+            {
+                // temp. printing for debugging
+                std::cout << "PMP addr update " << std::endl;
+
+                // In qemu, the rule is updated whenever
+                // pmpaddr is written (in other words,
+                // pmpcfg/addr are converted) into a rule
+
+                // PMP registers should only be modified in M mode
+                assert(readMiscRegNoEffect(MISCREG_PRV) == PRV_M);
```
powerjg, line 382: An assert makes sense. Should we also (instead) raise an exception? Or, would an permission exception be raised somewhere else?

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,19 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
...
+            // If we are not in the m-mode, check pmp
+            // permissions
```
powerjg, line 495: You should *always* check permissions, IIRC. I think you can revoke permission from m-mode (through the locked bit?). If you do this, it's permanent until reboot, but it's still something that's in the spec, IIRC.

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,19 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
...
+            if (pmode != PrivilegeMode::PRV_M) {
+                bool pass = walker->pmp->pmpCheck(req, mode);
+                if (!pass) {
```
powerjg, line 498: I don't love this variable name. I'd just do:
```suggestion
                if (!walker->pmp->pmpCheck(req, mode)) {
```

src/arch/riscv/pagetable_walker.cc
```diff
@@ -502,6 +513,20 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
...
+Fault
+Walker::WalkerState::createAddrfault(Addr vaddr, BaseTLB::Mode mode)
+{
+    ExceptionCode code;
+    if (mode == TLB::Read) {
+        code = ExceptionCode::LOAD_ACCESS;
+    } else if (mode == TLB::Write) {
+           code = ExceptionCode::STORE_ACCESS;
+    } else {
+            code = ExceptionCode::INST_ACCESS;
```
powerjg, line 525:
```suggestion
       code = ExceptionCode::STORE_ACCESS;
    } else {
        code = ExceptionCode::INST_ACCESS;
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+#include "math.h"
+#include "mem/request.hh"
+#include "params/PMP.hh"
+#include "sim/sim_object.hh"
+
+PMP::PMP(const Params &params) :
+SimObject(params), maxEntries(params.max_pmp)
+{
+
+    for (int i=0; i < maxEntries; i++) {
```
powerjg, line 42:
```suggestion
    for (int i=0; i < maxEntries; i++) {
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+#include "arch/generic/tlb.hh"
+#include "base/addr_range.hh"
+#include "base/types.hh"
+#include "math.h"
+#include "mem/request.hh"
+#include "params/PMP.hh"
+#include "sim/sim_object.hh"
+
+PMP::PMP(const Params &params) :
+SimObject(params), maxEntries(params.max_pmp)
```
powerjg, line 39:
```suggestion
PMP::PMP(const Params &params) :
    SimObject(params),
    maxEntries(params.max_pmp)
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+        AddrRange thisRange(0,-1);
+        entry->pmpAddr = thisRange;
+        entry->pmpCfg = 0;
+        pmpTable.push_back(entry);
+    }
+
+}
+
+// check if this region can be
+// accessed (else raise an exception)
```
powerjg, line 53: Should either be in the header file or in the body of the function where this happens.

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+
+    std::cout << "pmp: check" << std::endl;
+
+    // all pmp entries need to be looked from the lowest to
+    // the highest number
+
+    bool addrMatch = false;
+
+    int i;
+    for (i = 0; i < pmpTable.size(); i++) {
```
powerjg, line 66:
```suggestion
    for (int i = 0; i < pmpTable.size(); i++) {
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+        std::cout << "pmp: table check " << i << "   " << std::endl;
+        if (pmpTable[i]->pmpAddr.contains(req->getPaddr())){
+           //address matched
+            addrMatch = true;
+            break;
+        }
+    }
+
+    if (!addrMatch)
+        return false; //no pmp entry has a matching address
```
powerjg, line 76:
```suggestion
    // no pmp entry has a matching address
    if (!addrMatch) return false; 
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+bool
+PMP::pmpCheck(const RequestPtr &req, BaseTLB::Mode mode)
+{
+
+    std::cout << "pmp: check" << std::endl;
+
+    // all pmp entries need to be looked from the lowest to
+    // the highest number
+
+    bool addrMatch = false;
```
powerjg, line 63:
```suggestion
    int matched_index = -1
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+
+    int i;
+    for (i = 0; i < pmpTable.size(); i++) {
+        std::cout << "pmp: table check " << i << "   " << std::endl;
+        if (pmpTable[i]->pmpAddr.contains(req->getPaddr())){
+           //address matched
+            addrMatch = true;
```
powerjg, line 70:
```suggestion
            //address matched
            match_index = i;
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+    if (!addrMatch)
+        return false; //no pmp entry has a matching address
+
+
+    // else check the RWX permissions from the pmp entry
+
+    uint8_t allowedPrivs = PMP_READ | PMP_WRITE | PMP_EXEC;
+
+    // i is the index of pmp table which matched
+    allowedPrivs &= pmpTable[i]->pmpCfg;
```
powerjg, line 84:
```suggestion
    allowedPrivs &= pmpTable[match_index]->pmpCfg;
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+
+    if ((mode == BaseTLB::Mode::Read) &&
+            ((PMP_READ & allowedPrivs) == PMP_READ))
+        return true;
+    else if ((mode == BaseTLB::Mode::Write) &&
+            ((PMP_WRITE & allowedPrivs) == PMP_WRITE))
+        return true;
+    else if ((mode == BaseTLB::Mode::Execute) &&
+            ((PMP_WRITE & allowedPrivs) == PMP_EXEC))
+        return true;
```
powerjg, line 96:
```suggestion
    if ((mode == BaseTLB::Mode::Read) && (PMP_READ & allowedPrivs))
        return true;
    else if ((mode == BaseTLB::Mode::Write) && (PMP_WRITE & allowedPrivs))
        return true;
    else if ((mode == BaseTLB::Mode::Execute) && (PMP_WRITE & allowedPrivs))
        return true;
```
90% sure that's right :). The `allowed_privs & MASK` should be true iif it matches.

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+
+    if (!addrMatch)
+        return false; //no pmp entry has a matching address
+
+
+    // else check the RWX permissions from the pmp entry
+
+    uint8_t allowedPrivs = PMP_READ | PMP_WRITE | PMP_EXEC;
```
powerjg, line 81: Local variable should be snake_case

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+inline uint8_t
+PMP::pmpGetAField(uint8_t cfg)
+{
+    uint8_t a = cfg >> 3;
+    return a & 0x3;
+}
+
+
+void
+PMP::pmpUpdateCfg(uint32_t pmpIndex, uint8_t thisCfg)
```
powerjg, line 111: local variables (and function parameters) should be snake_case

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+    Addr prevAddr = 0;
+
+    if (pmpIndex >= 1) {
+
+        prevAddr = pmpTable[pmpIndex - 1]->pmpAddr.end();
+    }
+
+    switch (pmpGetAField(thisCfg)) {
+    case PMP_OFF:
+        {
```
powerjg, line 129: No need for curly braces (I think)

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+    uint8_t a = cfg >> 3;
+    return a & 0x3;
+}
+
+
+void
+PMP::pmpUpdateCfg(uint32_t pmpIndex, uint8_t thisCfg)
+{
+
+    std::cout << "pmp: update cfg" << std::endl;
```
powerjg, line 114: I think it's worth it to replace most of these with DPRINTF

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    switch (pmpGetAField(thisCfg)) {
+    case PMP_OFF:
+        {
+        AddrRange thisRange(0, -1);
+        pmpTable[pmpIndex]->pmpAddr = thisRange;
+        break;
+        }
+    case PMP_TOR:
```
powerjg, line 134: Could you add a comment after each of these describing what they mean? E.g., what is `TOR`?

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,94 @@
...
+{
+  public:
+
+    typedef PMPParams Params;
+
+    const Params &
+    params() const
+    {
+        return dynamic_cast<const Params &>(_params);
+    }
```
powerjg, line 53: There's a macro for this now. I forget what it is. See sim_object.hh

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,94 @@
...
+    const uint8_t PMP_EXEC  = 1 << 2;
+    const uint8_t PMP_LOCK  = 1 << 7;
+
+    // struct corresponding to a single
+    // PMP entry
+    typedef struct {
+        AddrRange pmpAddr;
+        Addr rawAddr;
+        uint8_t  pmpCfg;
+    } pmpEntry;
```
powerjg, line 82:
```suggestion
    } PmpEntry;
```

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,94 @@
...
+
+    std::vector<pmpEntry*> pmpTable;
```
powerjg, line 84:
```suggestion
    std::vector<PmpEntry> pmpTable;
```

Use `emplace` and don't use pointers when possible.

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,94 @@
...
+
+    bool pmpCheck(const RequestPtr &req, BaseTLB::Mode mode);
+    static inline uint8_t pmpGetAField(uint8_t cfg);
+    void pmpUpdateCfg(uint32_t pmpIndex, uint8_t thisCfg);
+    void pmpUpdateAddr(uint32_t pmpIndex, Addr thisAddr);
+    AddrRange pmpDecodeNapot(Addr a);
```
powerjg, line 90: Needs doxygen on this.

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,94 @@
...
+#include "params/PMP.hh"
+#include "sim/sim_object.hh"
+
+/**
+ * Place-holder for Doxygen comments on this class
+ */
+
+class PMP : public SimObject
+{
+  public:
```
powerjg, line 45: I think a lot of this can be private. Only expose what is necessary to the public.

src/arch/riscv/tlb.cc
```diff
@@ -262,6 +264,21 @@ TLB::createPagefault(Addr vaddr, Mode mode)
...
+Fault
+TLB::createAddrfault(Addr vaddr, Mode mode)
+{
+    ExceptionCode code;
+    if (mode == TLB::Read) {
+        code = ExceptionCode::LOAD_ACCESS;
+    } else if (mode == TLB::Write) {
+           code = ExceptionCode::STORE_ACCESS;
+    } else {
+            code = ExceptionCode::INST_ACCESS;
```
powerjg, line 277:
```suggestion
        code = ExceptionCode::STORE_ACCESS;
    } else {
        code = ExceptionCode::INST_ACCESS;
```
Spacing may be wrong.

src/arch/riscv/tlb.cc
```diff
@@ -366,6 +383,15 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
 
         if (!delayed && fault == NoFault) {
             pma->check(req);
+            // do pmp checks here if not in M mode
+            if (pmode != PrivilegeMode::PRV_M){
```
powerjg, line 387: Could you make the `pmpCheck` function return a fault (or `NoFault`)?

I see that the code below is duplicated here and in the walker. Can we move it to the `PMP` object?

powerjg (commented):
Getting closer, and it's looking good!

src/arch/riscv/isa.cc
```diff
@@ -331,6 +334,61 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+                // configured first, cfg or address regs!
+                // qemu seems to update the tables when
+                // pmp addr regs are written (with the assumption
+                // that cfg regs are already written)
+
+                // temp. printing for debugging
+                std::cout << "PMP cfg update " << std::endl;
+
+                int i;
+                for (i=0; i < sizeof(val); i++) {
```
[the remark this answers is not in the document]
ayaz91: fixed.

src/arch/riscv/isa.cc
ayaz91: updated.

src/arch/riscv/isa.cc
ayaz91: Yeah, the assert was supposed to be there temporarily. It was not obvious to me from the specs that if/what exception should be raised. I will look again into it.

src/arch/riscv/pagetable_walker.cc
ayaz91: I have updated the way we decide if we should check permissions. The update is based on a paragraph from the specs which mentions 3 conditions on which we should check permissions.

src/arch/riscv/pagetable_walker.cc
ayaz91: updated.

src/arch/riscv/pagetable_walker.cc
ayaz91: updated.

src/arch/riscv/pmp.cc
ayaz91: updated

src/arch/riscv/pmp.cc
ayaz91: updated.

src/arch/riscv/pmp.cc
ayaz91: moved to the function body.

src/arch/riscv/pmp.cc
ayaz91: fixed.

src/arch/riscv/pmp.cc
ayaz91: updated.

src/arch/riscv/pmp.cc
ayaz91: updated.

src/arch/riscv/pmp.cc
ayaz91: updated.

src/arch/riscv/pmp.cc
ayaz91: updated.

src/arch/riscv/pmp.cc
ayaz91: Yeah, this is correct (because the mask will only have 1 bit set at a time). fixed!

src/arch/riscv/pmp.cc
ayaz91: updated.

src/arch/riscv/pmp.cc
ayaz91: updated. I think I have updated everything in this changeset. But, I will go through this once again to make sure that everything is using the correct case

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,252 @@
...
+                ret = 0;
+                break;
+                }
+        }
+
+    }
+
+    // No PMP entry matched
+    if (ret == -1) {
+        if (pmode == RiscvISA::PrivilegeMode::PRV_M) {
```
powerjg, line 112: Move this check to the top. If we're in M, just return true.

If you do this, then you don't need `ret` or the `breaks`. Just return true/false if you have a match

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,252 @@
...
+        entry->pmpCfg = 0;
+        pmpTable.push_back(entry);
+    }
+
+    num_rules = 0;
+
+}
+
+bool
+PMP::pmpCheck(const RequestPtr &req, BaseTLB::Mode mode,
```
powerjg, line 58: This function is much more complicated than it needs to be. It could be done in ~15 lines.

src/arch/riscv/pmp.cc
powerjg: Still true...

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,252 @@
...
+        break;
+        }
+    default:
+        {
+        AddrRange thisRange(-1,-2);
+        pmpTable[pmp_index]->pmpAddr = thisRange;
+        break;
+        }
+    }
+        for (int i = 0; i < maxEntries; i++) {
```
powerjg, line 191:
```suggestion
    for (int i = 0; i < maxEntries; i++) {
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,252 @@
...
+    && (status.mpp != RiscvISA::PrivilegeMode::PRV_M));
+
+    return (cond1 || cond2 || cond3);
+}
+
+AddrRange
+PMP::pmpDecodeNapot(Addr a)
+{
+    if (a == -1) {
+      AddrRange thisRange(0u, -1);
```
powerjg, line 242: snake_case... why `0u`?

src/arch/riscv/pmp.hh
powerjg: ping

src/arch/riscv/pmp.hh
powerjg: ping

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,101 @@
...
+    typedef enum {
+        PMP_OFF,
+        PMP_TOR,
+        PMP_NA4,
+        PMP_NAPOT
+    } pmpAmatch;
+
+
+    // PMP range permissions
+    const uint8_t PMP_READ  =  1 << 0;
```
powerjg, line 72:
```suggestion
    const uint8_t PMP_READ  = 1 << 0;
```

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,101 @@
...
+    const uint8_t PMP_WRITE = 1 << 1;
+    const uint8_t PMP_EXEC  = 1 << 2;
+    const uint8_t PMP_LOCK  = 1 << 7;
+
+    int num_rules;
```
powerjg, line 77:
```suggestion
    int numRules = 0;
```
You can initialize in the header. It often cleans things up.

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,101 @@
...
+
+    // struct corresponding to a single
+    // PMP entry
+    typedef struct {
+        AddrRange pmpAddr;
+        Addr rawAddr;
+        uint8_t  pmpCfg;
```
powerjg, line 84: Give these defaults! Then you don't have to do it in the cpp code.

src/arch/riscv/pmp.hh
powerjg: Ping!

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,101 @@
...
+    } pmpEntry;
+
+    std::vector<pmpEntry*> pmpTable;
+
+    bool pmpCheck(const RequestPtr &req, BaseTLB::Mode mode,
+                RiscvISA::PrivilegeMode pmode);
```
powerjg, line 90:
```suggestion
    bool pmpCheck(const RequestPtr &req, BaseTLB::Mode mode,
                  RiscvISA::PrivilegeMode pmode);
```

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,101 @@
...
+    inline uint8_t pmpGetAField(uint8_t cfg);
+    void pmpUpdateCfg(uint32_t pmp_index, uint8_t this_cfg);
+    void pmpUpdateAddr(uint32_t pmp_index, Addr this_addr);
+    void pmpUpdateRule(uint32_t pmp_index);
+    bool shouldCheckPMP(RiscvISA::PrivilegeMode pmode,
+                BaseTLB::Mode mode, ThreadContext *tc);
+    inline AddrRange pmpDecodeNapot(Addr a);
```
powerjg, line 97: Which of these are actually public? All should have a comment describing what it does

src/arch/riscv/tlb.cc
```diff
@@ -262,6 +264,21 @@ TLB::createPagefault(Addr vaddr, Mode mode)
...
+// TODO: merge createPagefault and createAddrfault
+Fault
+TLB::createAddrfault(Addr vaddr, Mode mode)
+{
+    ExceptionCode code;
+    if (mode == TLB::Read) {
+        code = ExceptionCode::LOAD_ACCESS;
+    } else if (mode == TLB::Write) {
+        code = ExceptionCode::STORE_ACCESS;
+    } else {
```
powerjg, line 276: Is it true that Mode is one of these three? I suggest
```suggestion
    } else if (mode == TLB::Execute) {
```

src/arch/riscv/tlb.cc
```diff
@@ -366,6 +383,15 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
 
         if (!delayed && fault == NoFault) {
             pma->check(req);
+            // do pmp checks if any condition is met
+
+            if (pmp->shouldCheckPMP(pmode, mode, tc)){
```
powerjg, line 388: Can you hide this check in `pmpCheck`? In that function, can you first check if it's enabled/should be checked and return noFault in that case?

powerjg (commented):
More comments :). It's moving in a good direction!

src/arch/riscv/pagetable_walker.cc
ayaz91: Sorry, I am not sure how to check that in GitHub if file permission is changed?

src/arch/riscv/pmp.cc
ayaz91: fixed.

src/arch/riscv/pmp.cc
ayaz91: done

src/arch/riscv/pmp.cc
ayaz91: this code is not there any more.

src/arch/riscv/pmp.cc
ayaz91: Updated.

src/arch/riscv/pmp.cc
ayaz91: Still a todo to add debug printings

src/arch/riscv/pmp.cc
ayaz91: done

src/arch/riscv/pmp.hh
ayaz91: still todo!

src/arch/riscv/pmp.hh
ayaz91: done

src/arch/riscv/pmp.hh
ayaz91: I have added doxygen style comments. I think the syntax is correct, but I have not yet tried to generate documentation to check that

src/arch/riscv/pmp.hh
ayaz91: updated

src/arch/riscv/tlb.cc
ayaz91: done

src/arch/riscv/tlb.cc
ayaz91: done. These functions are moved to the PMP class

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,21 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
...
             req->setPaddr(paddr);
             walker->pma->check(req);
+
+            // do pmp checks if any condition is met
+            if (walker->pmp->shouldCheckPMP(pmode, mode, tc)){
+                // raise exception if checks do not pass
+                if (!walker->pmp->pmpCheck(req, mode, pmode)) {
+                    timingFault = createAddrfault(req->getVaddr(), mode);
+                }
+                else{
```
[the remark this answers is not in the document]
ayaz91: fixed.

src/arch/riscv/pagetable_walker.cc
```diff
@@ -490,8 +490,21 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
             Addr paddr = walker->tlb->translateWithTLB(vaddr, satp.asid, mode);
             req->setPaddr(paddr);
             walker->pma->check(req);
+
+            // do pmp checks if any condition is met
+            if (walker->pmp->shouldCheckPMP(pmode, mode, tc)){
```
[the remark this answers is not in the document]
ayaz91: Sorry, trying to be more careful!

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,252 @@
...
+#include "mem/request.hh"
+#include "params/PMP.hh"
+#include "sim/sim_object.hh"
+
+PMP::PMP(const Params &params) :
+    SimObject(params),
+    maxEntries(params.max_pmp)
+{
+    for (int i=0; i < maxEntries; i++) {
+        pmpEntry* entry = new pmpEntry;
```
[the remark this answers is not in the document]
ayaz91: done

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,252 @@
...
+        AddrRange thisRange(-1,-2);
+        entry->pmpAddr = thisRange;
+        entry->pmpCfg = 0;
+        pmpTable.push_back(entry);
+    }
+
+    num_rules = 0;
```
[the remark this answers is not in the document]
ayaz91: done

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,252 @@
...
+
+    for (int i = 0; i < pmpTable.size(); i++) {
+        //std::cout << "pmp: table check " << i << "   " << std::endl;
+        if (pmpTable[i]->pmpAddr.contains(req->getPaddr())){
+           //address matched
+            match_index = i;
+        }
+
+        if ((PMP_OFF != pmpGetAField(pmpTable[match_index]->pmpCfg))
+                                            && (match_index > -1)){
```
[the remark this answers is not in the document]
ayaz91: updated

src/arch/riscv/pmp.cc
ayaz91: done

src/arch/riscv/pmp.cc
ayaz91: I think this function is much simpler now.

src/arch/riscv/pmp.cc
ayaz91: updated

src/arch/riscv/pmp.cc
ayaz91: fixed.

src/arch/riscv/pmp.hh
ayaz91: fixed.

src/arch/riscv/pmp.hh
ayaz91: initialized in the initializer list.

src/arch/riscv/pmp.hh
ayaz91: done

src/arch/riscv/pmp.hh
ayaz91: done

src/arch/riscv/pmp.hh
ayaz91: Added comments and updated private/public visibility.

src/arch/riscv/tlb.cc
ayaz91: Yes, Mode is an enum with three possible values read, write, execute

src/arch/riscv/tlb.cc
ayaz91: done

src/arch/riscv/pagetable_walker.cc
powerjg: Screenshot_20210319_074336

Do you see where it says "100644->100755"? That's saying the permissions changed. 

I think you can use `git ls-files` to see it in your local repo.

src/arch/riscv/pagetable_walker.hh
```diff
@@ -163,11 +164,13 @@ namespace RiscvISA
         Port &getPort(const std::string &if_name,
                       PortID idx=InvalidPortID) override;
 
+
```
powerjg, line 167: extra newline?

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,223 @@
...
+
+            // i is the index of pmp table which matched
+            allowed_privs &= pmpTable[match_index].pmpCfg;
+
+            // mode: R = 0, W = 1, X = 2
+            uint8_t mode_mask = 1 << mode;
+            if ((allowed_privs & mode_mask) > 0)
+                return NoFault;
+            else
+                return createAddrfault(req->getVaddr(), mode);
```
powerjg, line 89:
```suggestion
            if ((allowed_privs & mode_mask) > 0) {
                return NoFault;
            }
            else {
                return createAddrfault(req->getVaddr(), mode);
            }
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,223 @@
...
+        if ((PMP_OFF != pmpGetAField(pmpTable[match_index].pmpCfg))
+                                            && (match_index > -1)) {
+            // check the RWX permissions from the pmp entry
+            uint8_t allowed_privs = PMP_READ | PMP_WRITE | PMP_EXEC;
+
+            // i is the index of pmp table which matched
+            allowed_privs &= pmpTable[match_index].pmpCfg;
+
+            // mode: R = 0, W = 1, X = 2
+            uint8_t mode_mask = 1 << mode;
```
powerjg, line 85: I'm not sure you can count on TLB::Mode to be 0, 1, or 2. Even though it increases the code, I think the full checks would be safer (e.g., `if (mode == TLB::Read)`). Let me know if I'm wrong.

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,223 @@
...
+
+    // An access should be successful if there are
+    // no rules defined yet or we are in M mode (based
+    // on specs v1.10)
+    if (numRules == 0 || (pmode == RiscvISA::PrivilegeMode::PRV_M))
+        return NoFault;
+
+    // all pmp entries need to be looked from the lowest to
+    // the highest number
+    int match_index = -1;
```
powerjg, line 69: I think this is needed because for range entries you actually need two PMP registers, right? I think it could be worth a comment about this.

src/arch/riscv/pmp.cc
powerjg: Yes! Much improved! Thanks!

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    pmpTable[pmp_index].pmpAddr = this_range;
+
+    for (int i = 0; i < maxEntries; i++) {
+      const uint8_t a_field =
+        pmpGetAField(pmpTable[i].pmpCfg);
+      if (PMP_OFF != a_field) {
+        numRules++;
+      }
```
powerjg, line 173: Switched back to two spaces for some reason?

```suggestion
        const uint8_t a_field =
            pmpGetAField(pmpTable[i].pmpCfg);
        if (PMP_OFF != a_field) {
            numRules++;
        }
```

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,223 @@
...
+        prevAddr = pmpTable[pmp_index - 1].rawAddr;
+    }
+
+    Addr this_addr = pmpTable[pmp_index].rawAddr;
+    uint8_t this_cfg = pmpTable[pmp_index].pmpCfg;
+    AddrRange this_range;
+
+    switch (pmpGetAField(this_cfg)) {
+      // checking the address matching mode of pmp entry
+      case PMP_OFF:
```
powerjg, line 146: I *think* these case statements should be at the same indentation as the switch. I could be wrong, though. I thought that the `public/private` keywords were the only thing we had at 2 spaces indented.

src/arch/riscv/pmp.hh
```diff
@@ -0,0 +1,194 @@
...
+    /** pmpcfg address range locked mask */
+    const uint8_t PMP_LOCK = 1 << 7;
+
+    /** variable to keep track of active number of rules any time */
+    int numRules;
+
+    /** single pmp entry struct*/
+    typedef struct {
+        /** addr range corresponding to a single pmp entry */
+        AddrRange pmpAddr = AddrRange(-1, -2);
```
powerjg, line 100: I suggest initializing this to `AddrRange(0, 0)` I'm not sure what a negative address means. I think it's a uint64_t, so these are actually just big numbers. (Or, they could be ints cast to uint64_t in which case it's 0xFFFFFFFF).

(0,0) is an empty range, so it won't match anything

powerjg (commented):
It keeps improving! Thanks, Ayaz!

src/arch/riscv/pagetable_walker.hh
ayaz91: fixed. somehow left this line when I deleted an unwanted function declaration (that I had previously added)

src/arch/riscv/pmp.cc
ayaz91: fixed.

src/arch/riscv/pmp.cc
ayaz91: I think it can be a problem only if in the future the encoding of TLB::Mode is changed to something else. Otherwise, I can't think of a case where the use of mode_mask will be a problem. However, using mode == TLB::Read is definitely more readable and easy to understand. I can get rid of mode_mask if you suggest!

src/arch/riscv/pmp.cc
ayaz91: Assuming that you are referring to TOR mode, the pmpUpdateRule takes care of this and creates a single address range for that particular pmp entry. In pmpCheck match_index is needed just to keep track of which pmp entry the address matched with so that the index can be used to access permission bits from that pmp entry.

src/arch/riscv/pmp.cc
ayaz91: fixed.

src/arch/riscv/pmp.cc
ayaz91: I couldn't find any discussion on this in the style guide and looking into some other source files (e.g. isa.cc) it seems like there are 2 spaces indented cases that do not seem to have keywords in the case statement. Should I keep it as it is or just remove these spaces?

src/arch/riscv/pmp.hh
ayaz91: fixed to (0,0). There was no need to have negative addresses. I think I was just trying to debug something and though that I am having unintended matches for address 0, so changed the defaults to negative addresses and did not revert them later.

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,232 @@
...
+
+PMP::PMP(const Params &params) :
+    SimObject(params),
+    maxEntries(params.max_pmp),
+    numRules(0)
+{
+    for (int i=0; i < maxEntries; i++) {
+        PmpEntry entry;
+        pmpTable.emplace_back(entry);
+    }
```
powerjg, line 50:
```suggestion
    pmpTable.resize(maxEntries)
```

src/arch/riscv/pmp.cc
powerjg: Yeah, I'd do that

src/arch/riscv/pmp.cc
```diff
@@ -0,0 +1,232 @@
...
+
+AddrRange
+PMP::pmpDecodeNapot(Addr pmpaddr)
+{
+    if (pmpaddr == -1) {
+      AddrRange this_range(0, -1);
+      return this_range;
+    } else {
+        uint64_t t1 = ctz64(~pmpaddr);
+        uint64_t range = std::pow(2,t1+3);
```
powerjg, line 227:
```suggestion
        uint64_t range = 1 << (t1+3);
```
Optional

src/arch/riscv/pmp.cc
```diff
@@ -224,9 +233,12 @@ PMP::pmpDecodeNapot(Addr pmpaddr)
...
+        uint64_t range = (std::pow(2,t1+3))-1;
+
+        // pmpaddr reg encodes bits 55-2 of a
+        // 56 bit physical address for RV64
+        uint64_t base = mbits(pmpaddr, 63, t1) << 2;
         AddrRange this_range(base, base+range);
         return this_range;
     }
-}
+}
```
powerjg, line 244: Lost the newline at the end of the file.

powerjg (commented):
Time to push to gerrit?

Before you do that, make sure you fix the permissions :)

powerjg:
I think this can be closed, right?

PR #218, opened 2021-03-10
Jsonloader

yllekn:


src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,55 @@
...
+                                unit=d.get('unit'),
+                                description=d.get('description'),
+                                datatype=d.get('datatype')
+                               )
+        elif d['type'] == 'Group':
+            return Group(**d)
+        elif d['type'] == 'Vector':
+            return Group(**d)
+    else:
+        return SimStat(**d)
```
powerjg, line 55: Can you set your editor to create newlines at the end of files?

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,55 @@
...
+                                max=d.get('max'),
+                                sumSquared=d.get('sum_squared'),
+                                unit=d.get('unit'),
+                                description=d.get('description'),
+                                datatype=d.get('datatype')
+                               )
+        elif d['type'] == 'Group':
+            return Group(**d)
+        elif d['type'] == 'Vector':
+            return Group(**d)
```
powerjg, line 53: Why not 
```suggestion
            return Vector(**d)
```

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,55 @@
...
+        simstat_obj = json.load(f, object_hook=__json_to_simstat)
+
+    return simstat_obj
+
+def __json_to_simstat(d: dict) -> SimStat:
+    print(d)
+    if 'type' in d:
+        if d['type'] == 'Scalar':
+            return Scalar(
+                          value=d.get('value'),
```
powerjg, line 18: Why `d.get` and not `d['value']`?

powerjg (commented):
LGTM! Two small things

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,55 @@
+from .simstat import SimStat
```
powerjg, line 1: Needs a copyright

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,55 @@
...
+from .statistic import Scalar, Distribution, Accumulator
+from .group import Group, Vector
+import json
+
+def json_loader(path: str) -> SimStat:
```
powerjg, line 6: maybe just call this `load`? Is there a way it could have the *same* interface as `json.load`?

src/python/m5/ext/pystats/jsonloader.py
yllekn: Some of these statistics have optional data members, and .get() supports returning NoneType (rather than throwing an error). I made all of them .get() for the sake of continuity, but I could differentiate between optional (d.get()) and non-optional data members (d['value']) if you like

src/python/m5/ext/pystats/jsonloader.py
yllekn: Maybe we don't even need this wrapper function then. The user can just call 
> simstat_obj = json.load(f, object_hook=__json_to_simstat)
like in line 9.

powerjg: I like the idea of hiding the `object_hook` from "regular" users.

src/python/m5/ext/pystats/jsonloader.py
powerjg: I see! Would it work to just do the following?

```python
return Scalar(**d)
```

yllekn: I don't think so. These statistics don't take **kwargs as an argument.

yllekn: Let me try it and see what happens though.

src/python/m5/ext/pystats/jsonloader.py
yllekn: Ok. I will look into subclassing the JSONDecoder then.

src/python/m5/ext/pystats/jsonloader.py
powerjg: All python functions take a dictionary as a parameter :smile:

See, for instance, https://book.pythontips.com/en/latest/args_and_kwargs.html

src/python/m5/ext/pystats/jsonloader.py
yllekn: Would something like json.loads(f, cls=JSONLoader) be acceptable, where JSONLoader is a subclass of JSONDecoder with object_hooks overwritten? I am trying to think of a way that we could hide the underlying "JSONLoader" stuff from user entirely

src/python/m5/ext/pystats/jsonloader.py
yllekn: I believe you... it works but it requires removing the "type" from the dictionary d, just because of the way we've defined the constructor for statistics and Vectors. I'll push an updated version for you soon to check out

powerjg: Makes sense. One thing we could to is add `**kwargs` as the final parameter to the stats objects to allow other kwargs.

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -1,55 +1,72 @@
...
 import json
 
-def json_loader(path: str) -> SimStat:
-    
+class JSONLoader(json.JSONDecoder):
+    """
+    Subclass of JSONDecoder that overrides 'object_hook'. Converts JSON object
+    into a SimStat object.
+
+    Usage
```
yllekn, line 37: @powerjg What do you think of this interface?

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -1,55 +1,72 @@
...
+    ```
+    """
+
+    def __init__(self):
+        json.JSONDecoder.__init__(self, object_hook=self.__json_to_simstat)
+
+    def __json_to_simstat(self, d: dict) -> SimStat:
+        if 'type' in d:
+            if d['type'] == 'Scalar':
+                d.pop('type', None)
```
yllekn, line 53: Not ideal...

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,72 @@
...
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from .simstat import SimStat
+from .statistic import Scalar, Distribution, Accumulator
+from .group import Group, Vector
+import json
+
+class JSONLoader(json.JSONDecoder):
```
powerjg, line 32: This is *perfect*!

I suggest also adding a function `load` so you can use the following:
```
with open('stats.json') as f:
  pystats.jsonloader.load(f)
```
This load function should be *similar* (though not the same) as the `json.load` function.

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,72 @@
...
+    ```
+    """
+
+    def __init__(self):
+        json.JSONDecoder.__init__(self, object_hook=self.__json_to_simstat)
+
+    def __json_to_simstat(self, d: dict) -> SimStat:
+        if 'type' in d:
+            if d['type'] == 'Scalar':
+                d.pop('type', None)
```
powerjg, line 53: I suggest dropping these. Instead, I think that `Scalar`, `Vector`, etc. should be modified something like below...

```
class Scalar(Statistic):
    """
    A scalar Python statistic type.
    """

    value: Union[float, int]

    def __init__(self, value: Any,
                 unit: Optional[str] = None,
                 description: Optional[str] = None,
                 datatype: Optional[StorageType] = None,
                 **kwargs = None):
        super(Scalar, self).__init__(
                                     value=value,
                                     type="Scalar",
                                     unit=unit,
                                     description=description,
                                     datatype=datatype,
                                     **kwargs
                                    )
```

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -1,55 +1,72 @@
...
-        elif d['type'] == 'Group':
-            return Group(**d)
-        elif d['type'] == 'Vector':
-            return Group(**d)
-    else:
-        return SimStat(**d)
\ No newline at end of file
+    simstat_object = json.load(f, cls=JSONLoader)
+    ```
+    """
```
yllekn: @powerjg What do you think of this interface, where caller must specify 'cls' argument?

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -1,55 +1,72 @@
...
+            elif d['type'] == 'Accumulator':
+                d.pop('type', None)
+                return Accumulator(**d)
+
+            elif d['type'] == 'Group':
+                return Group(**d)
+
+            elif d['type'] == 'Vector':
+                d.pop('type', None)
+                d.pop('time_conversion', None)
```
yllekn, line 69: d.pop() is not ideal, but it works

src/python/m5/ext/pystats/jsonloader.py
yllekn: Sorry, not sure why this comment is so weirdly formatted. I'm referring to the 'Usage'

powerjg (commented):
Looks great!

src/python/m5/ext/pystats/jsonloader.py
yllekn: Does this solve the problem? (I will try this in just a second to find out.) I don't think the issue was that the constructor wouldn't take additional arguments, but it was because it was receiving two of the same argument, e.g. 'type' was received twice?

src/python/m5/ext/pystats/jsonloader.py
yllekn: So just like a wrapper function for json.load?

powerjg: Yeah, exactly!

src/python/m5/ext/pystats/jsonloader.py
powerjg: I *think* so. I meant to tag @BobbyRBruce to see what he thinks about this as well. 

I could be wrong, though

yllekn: I tried it, and ran into the same error from before I added the d.pop() statements:

`  File "build/NULL/python/m5/ext/pystats/jsonloader.py", line 53, in __json_to_simstat
  File "build/NULL/python/m5/ext/pystats/statistic.py", line 72, in __init__
TypeError: __init__() got multiple values for keyword argument 'type'`

powerjg: Gotcha... Oh! I see... 
```
( value=value, type="Scalar",
...
```

I'm good with this change as it is, then :)

src/python/m5/ext/pystats/jsonloader.py
powerjg: @BobbyRBruce, why does vector not have a time conversion, but everything else does...?

BobbyRBruce: It does have time_conversation. Vector inherets from "Group" which contains it.

powerjg: So, why `pop('time_conversion')`, @yllekn?

yllekn: The constructor for Vector has 'time_conversion = None' in it. I'm not sure what the intent was, or if it was meant to be an optional argument, but it's the same issue as before with 'type'-- setting the property 'time_conversion' twice.

yllekn: In other words, I get that same 'multiple values' error.

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -70,3 +70,20 @@ def __json_to_simstat(self, d: dict) -> SimStat:
...
+    ```
+    import m5.ext.pystats as pystats
+
+    with open(path) as f:
+        pystats.jsonloader.load(f)
+    ```
+    """
+
+    simstat_object = json.load(json_file, cls=JSONLoader)
+    return simstat_object
```
yllekn, line 89: @powerjg

powerjg:
Did you do a force push to the jsonloader branch?

@BobbyRBruce, could you do a quick review before we push to gerrit?

yllekn:
@powerjg Sorry I forgot that you mentioned that. It's done now.

src/python/m5/ext/pystats/simstat.py
```diff
@@ -42,10 +42,10 @@ class SimStat(JsonSerializable):
     simulated_begin_time: Optional[Union[int, float]]
     simulated_end_time: Optional[Union[int, float]]
 
-    def __init__(self, creation_time: Optional[datetime],
-                 time_conversion: Optional[TimeConversion],
-                 simulated_begin_time: Optional[Union[int, float]],
-                 simulated_end_time: Optional[Union[int, float]],
+    def __init__(self, creation_time: Optional[datetime] = None,
```
powerjg, line 45: This should be a separate commit:
"python: Fix optional parameters with default None"

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) 2020 The Regents of The University of California
```
powerjg, line 1:
```suggestion
# Copyright (c) 2021 The Regents of The University of California
```

powerjg (commented):
This is good with me. I'd like @BobbyRBruce to take a quick look, too.

My comments are now about how best to push to gerrit, not on the changes themselves ;).

src/python/m5/ext/pystats/__init__.py
```diff
@@ -30,6 +30,7 @@
 from .statistic import Statistic
 from .storagetype import StorageType
 from .timeconversion import TimeConversion
+from .jsonloader import JSONLoader
```
BobbyRBruce, line 33: It's not a big problem but I do format this as "Json" for other classes. E.g., "JsonSerializable"

src/python/m5/ext/pystats/simstat.py
BobbyRBruce: Agreed.

src/python/m5/stats/gem5stats.py
```diff
@@ -230,7 +230,7 @@ def _prepare_stats(group: _m5.stats.Group):
     for stat in group.getStats():
         stat.prepare()
 
-    for child in getStatGroups().values():
+    for child in group.getStatGroups().values():
```
BobbyRBruce, line 233: Just be careful here. This has already been fixed: https://gem5-review.googlesource.com/c/public/gem5/+/42704 (this will be merged into develop this afternoon: https://gem5-review.googlesource.com/c/public/gem5/+/43265).

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,89 @@
...
+
+            elif d['type'] == 'Group':
+                return Group(**d)
+
+            elif d['type'] == 'Vector':
+                d.pop('type', None)
+                d.pop('time_conversion', None)
+                return Vector(d)
+        else:
+            return SimStat(**d)
```
BobbyRBruce, line 72: Can we have some fail-cases in `__json_to_simstat` function? I'm thinking if someone adds a new stat type, this function should throw an "unrecognized type" error.

I'm also a bit uncomfortable that the definition of a SimStat is anything that doesn't have a "type". Is there a way we can improve this somehow? Maybe give SimStat a type of "simstat" ? It would be worrying if someone extended the Python Stats with something that doesn't have a type and they are cast to SimStats (I guess this would throw an error somewhere but it might be hard to diagnose?).

src/python/m5/ext/pystats/jsonloader.py
```diff
@@ -0,0 +1,89 @@
...
+    Subclass of JSONDecoder that overrides 'object_hook'. Converts JSON object
+    into a SimStat object.
+
+    Usage
+    -----
+    ```
+    from m5.ext.pystats.jsonloader import JSONLoader
+
+    with open(path) as f:
+        simstat_object = json.load(f, cls=JSONLoader)
```
BobbyRBruce, line 43: I got a little confused here due to the function at https://github.com/darchr/gem5/pull/218/files#diff-7d72dc1dd99fe1127f61d32392d0943663f4187f7249fcfd68c667bdae10530fR74 . Perhaps it's best to mention here that a helper function is provided to use this class in a cleaner way.

BobbyRBruce (commented):
Well done Kelly! Though i do have some comments, I'm mostly happy with these changes. I'm surprised how clearly the Hooks works. If there are next steps to this I'd recommend:

- Tests which run in our TestLib Framework. I think, as a minimum, check that the SimStats from a basic simulation can be translated to and from JSON while preserving all the data (I would actually run a small simulation for this so we know the PyBindings are still working and the translation is still good). 

- Someone needs to update Jason's schema: https://github.com/gem5/stats-schema. It's out of date due to some of the latest changes.

src/python/m5/ext/pystats/jsonloader.py
yllekn: lol

src/python/m5/ext/pystats/simstat.py
yllekn: I can try to figure out how to do this

src/python/m5/stats/gem5stats.py
yllekn: Hm. How to deal with this then? Wait until this is merged with develop and then rebase or something?

src/python/m5/ext/pystats/jsonloader.py
yllekn: Maybe @powerjg could weigh in on this

src/python/m5/ext/pystats/jsonloader.py
yllekn: resolved [here](https://github.com/darchr/gem5/compare/d974742952a4acfd9a5711f35d25cd42e84d0e70..bf98acaa9d8631db3a97f15eb4bd06524d690d68)

yllekn: resolved [here](https://github.com/darchr/gem5/compare/d974742952a4acfd9a5711f35d25cd42e84d0e70..bf98acaa9d8631db3a97f15eb4bd06524d690d68)

src/python/m5/ext/pystats/__init__.py
yllekn: resolved [here](https://github.com/darchr/gem5/compare/d974742952a4acfd9a5711f35d25cd42e84d0e70..bf98acaa9d8631db3a97f15eb4bd06524d690d68)

src/python/m5/ext/pystats/jsonloader.py
yllekn: resolved [here](https://github.com/darchr/gem5/compare/d974742952a4acfd9a5711f35d25cd42e84d0e70..bf98acaa9d8631db3a97f15eb4bd06524d690d68)

src/python/m5/ext/pystats/simstat.py
yllekn: resolved [here](https://github.com/darchr/gem5/compare/d974742952a4acfd9a5711f35d25cd42e84d0e70..bf98acaa9d8631db3a97f15eb4bd06524d690d68)

yllekn:
I rebased the branch to split a commit for Optional parameters and make some other changes. I'll get to testing per Bobby's suggestion soon. Would appreciate some feedback on the above unresolved issues @powerjg @BobbyRBruce

src/python/m5/ext/pystats/jsonloader.py
powerjg: I think this is fine for now. I don't know how better to do this if we're deserializing from json. I think we're going to want to provide simplified output formats (i.e., without the type) and then we'll have to make even more "guesses" with the deserializer.

If you know better ways to solve this problem, Bobby, I'm all ears.

src/python/m5/stats/gem5stats.py
powerjg: Exactly :)

powerjg (commented):
I think one more rebase is in order, but then you can push to gerrit. I think there should be two commits: 1. fix the missing nones 2. Add the json loader.

src/python/m5/stats/gem5stats.py
yllekn: @powerjg Doesn't look like this change is reflected in the github repo, am I supposed to cherry-pick this commit from googlesource or something like that?

PR #222, opened 2021-05-19
Adding traffic generators to component library.

mahyarsamani:
…_core.

This patch adds scripts that define an abstract core for traffic generators
and a simple core that is specific to PyTrafficGen in gem5. The class
AbstractGeneratorCore has an extended interface for set_traffic and
start_traffic functions.


components_library/processors/abstract_generator_core.py
```diff
@@ -0,0 +1,68 @@
...
+from .abstract_core import AbstractCore
+
+class AbstractGeneratorCore(AbstractCore):
+    def __init__(self, cpu_type: CPUTypes):
+        super(AbstractGeneratorCore, self).__init__(cpu_type)
+        assert cpu_type == CPUTypes.PYGEN or cpu_type == CPUTypes.GUPSGEN
+        self.setup_dummy_generator()
+
+    def connect_icache(self, port: Port) -> None:
+        self.dummy_generator.port = port
```
powerjg, line 40: I wonder what happens if you just do nothing (i.e., `pass`) here. Rather than creating an empty traffic generator, it would be nice to have a way to just not connect the port or to connect it to a "Null" port.

components_library/processors/cpu_types.py
```diff
@@ -32,3 +32,5 @@ class CPUTypes(Enum):
     KVM = 2
     O3 = 3
     TIMING = 4
+    PYGEN = 5
```
powerjg, line 35: This isn't scalable... we're should try to find a better solution here. @BobbyRBruce

components_library/processors/py_generator_core.py
```diff
@@ -0,0 +1,43 @@
...
+
+from .abstract_generator_core import AbstractGeneratorCore
+
+class PyGeneratorCore(AbstractGeneratorCore):
+    def __init__(self):
+        super(PyGeneratorCore, self).__init__()
+        self.main_generator = PyTrafficGen()
+
+    def connect_dcache(self, port: Port) -> None:
+        self.main_generator.port = port
```
powerjg, line 37: @BobbyRBruce, I think we should not use `=` to connect ports. I *think* we can use `port.connect(self.main_generator, port)` or something close to that here. We can do similar things in other parts of the codebase.

I think that `connect`'s semantics are much clearer.

components_library/processors/linear_generator_core.py
```diff
@@ -0,0 +1,36 @@
...
+        max_period = period
+        yield self.main_generator.createLinear(
+            duration,
+            min_addr,
+            max_addr,
+            block_size,
+            min_period,
+            max_period,
+            rd_perc,
+            data_limit,
```
powerjg, line 31:
```suggestion
        yield self.main_generator.createLinear(
            duration = 100000000
        block_size = 64
        min_addr = 0
        max_addr = 16384
        rd_perc = 100
        data_limit = 0
        period = int(float(block_size / rate) * float(1e12 / 1073741824))
        min_period = period
        max_period = period
```

powerjg (commented):
Cool! And this works?! This is perfect :smile:

components_library/memory/single_channel.py
```diff
@@ -104,3 +107,71 @@ def SingleChannelDDR3_2133(size: Optional[str] = None) -> AbstractMemorySystem:
...
+
+def SingleChannelHBM(size: Optional[str] = None) -> AbstractMemorySystem:
+    return SingleChannelMemory(HBM_1000_4H_1x128, size)
+
+
+from .DS3.DS3Ctrl import DS3MemCtrl
+
+
+class DS3SingleChannel(AbstractMemorySystem):
+    def __init__(self, mem_type: str, size: Optional[str]):
```
powerjg, line 133: Please split all of this out into another python file. E.g., dramsim_3.py

So, you could then say
```python
from dramsim_3 import SingleChannelDDR3_1600
```
It would even be OK (preferable?) to have the same class names!

powerjg (commented):
general comment: rename DS3 -> dramsim_3

components_library/processors/complex_generator.py
```diff
@@ -0,0 +1,104 @@
...
+
+class ComplexGenerator(AbstractProcessor):
+    """
+    A SimpeProcessor contains a number of cores of a a single CPUType.
+    """
+
+    def __init__(
+        self,
+        num_cores: int,
+    ) -> None:
```
powerjg, line 43: Does this not fit on one line?

components_library/processors/complex_generator.py
```diff
@@ -0,0 +1,104 @@
...
+        super(ComplexGenerator, self).__init__(
+            cores=self._create_cores(
+                num_cores=num_cores,
+            )
+        )
+
+    def _create_cores(
+        self,
+        num_cores,
+    ):
```
powerjg, line 53: One line

components_library/processors/complex_generator.py
```diff
@@ -0,0 +1,104 @@
...
+        return [ComplexGeneratorCore() for i in range(num_cores)]
+
+    def incorporate_processor(self, board: AbstractBoard) -> None:
+        # TODO: Shouldn't we do this with a setter function?
+        board.mem_mode = "timing"
```
powerjg, line 58: @BobbyRBruce, looks like we need to do this "TODO" sooner rather than later if the code is being copied around.

components_library/processors/complex_generator_core.py
```diff
@@ -0,0 +1,116 @@
...
+    def add_linear(
+        self,
+        duration="1ms",
+        rate="100GB/s",
+        block_size=64,
+        min_addr=0,
+        max_addr=32768,
+        rd_perc=100,
+        data_limit=0,
+    ):
```
powerjg, line 26: Let's talk about what is good to have a "reasonable default" vs required. Some of these you can't guess for the user.

components_library/processors/complex_generator_core.py
```diff
@@ -0,0 +1,116 @@
...
+        duration = int(toLatency(duration) * 1e12)
```
powerjg, line 27: You cannot assume a tick is 1 ps

components_library/processors/complex_generator_core.py
```diff
@@ -0,0 +1,116 @@
+from m5.util.convert import toLatency, toMemoryBandwidth
+
+from components_library.processors.cpu_types import CPUTypes
+from components_library.processors.py_generator_core import PyGeneratorCore
+
+
+from .py_generator_core import PyGeneratorCore
+
+
+class ComplexGeneratorCore(PyGeneratorCore):
```
powerjg, line 10: This needs tons of comments. I'm an expert at this stuff and I don't understand how I'm supposed to use any function in this file.

components_library/processors/complex_generator.py
```diff
@@ -0,0 +1,104 @@
...
+
+from .abstract_processor import AbstractProcessor
+from ..boards.abstract_board import AbstractBoard
+
+from typing import List
+
+
+class ComplexGenerator(AbstractProcessor):
+    """
+    A SimpeProcessor contains a number of cores of a a single CPUType.
```
powerjg, line 37: This isn't even true... Please add comments explaining what this processor is and how to use it.

components_library/processors/cpu_types.py
```diff
@@ -32,3 +32,6 @@ class CPUTypes(Enum):
     KVM = 2
     O3 = 3
     TIMING = 4
+    LINGEN = 5
+    RNDGEN = 6
+    CMXGEN = 7
```
powerjg, line 37: I am strongly against this kind of change. We have to find a better way.

components_library/processors/linear_generator.py
```diff
@@ -0,0 +1,99 @@
...
+
+from .abstract_processor import AbstractProcessor
+from ..boards.abstract_board import AbstractBoard
+
+from typing import List
+
+
+class LinearGenerator(AbstractProcessor):
+    """
+    A SimpeProcessor contains a number of cores of a a single CPUType.
```
powerjg, line 37: Fix documentation

powerjg (commented):
Great start! It needs some documentation so people know how to use it, though :).

Also, we need to talk about the interface to the complex generator

components_library/processors/cpu_types.py
BobbyRBruce: Is it necessary to specify each of the random cores in this way? Could we not have something like "MISC" core for all these? A kinda "catch-all" for cases like this?

Alternatively we can abstract our core hierarchy a bit more:

"AbstractNormalCore" (accepts CPU type as an argument) and "AbstractSpecialCore" (does not have a CPU type), both of which inherit from "AbstractCore".

Another possibility: Multiple inheritance. SimpleCores inherit from both "AbstractCore" and "gem5Core", gem5Core is the one that enforces the CPUTypes.

components_library/processors/complex_generator.py
```diff
@@ -0,0 +1,104 @@
...
+    def add_linear(
+        self,
+        duration="1ms",
+        rate="100GB/s",
+        block_size=64,
+        min_addr=0,
+        max_addr=32768,
+        rd_perc=100,
+        data_limit=0,
+    ):
```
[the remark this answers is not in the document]
BobbyRBruce: Ditto for almost every function you've created in this pull-request :)

components_library/processors/gups_generator.py
```diff
@@ -0,0 +1,71 @@
...
+from ..boards.abstract_board import AbstractBoard
+
+from typing import List
+
+
+class GUPSGenerator(AbstractProcessor):
+    def __init__(
+        self,
+        num_cores=1,
+        mode="ep",
```
BobbyRBruce, line 42: "mode", could we be more clever here and create an enum? I'm really not a fan of using strings to communicate discrete information like this.

components_library/processors/gups_generator.py
```diff
@@ -0,0 +1,71 @@
...
+        start_addr=0,
+        mem_size="32MB",
```
BobbyRBruce, line 44: Note to self: I don't like how Sizes are communicated strings. I'd like to wrap this in a type one day.

BobbyRBruce: Maybe even Addresses too...

components_library/processors/py_generator_core.py
BobbyRBruce: I agree.

components_library/processors/py_generator_core.py
```diff
@@ -0,0 +1,48 @@
...
+        super(PyGeneratorCore, self).__init__(cpu_type=cpu_type)
+        self.main_generator = PyTrafficGen()
+
+    def connect_dcache(self, port: Port) -> None:
+        self.main_generator.port = port
+
+    def create_traffic(self):
+        raise NotImplementedError
+
+    def set_traffic(self):
```
BobbyRBruce, line 44: Please annotate these abstract methods as abstract.

components_library/processors/random_generator_core.py
```diff
@@ -0,0 +1,48 @@
...
+        self._duration = int(toLatency(duration) * 1e12)
+        self._rate = toMemoryBandwidth(rate)
+        self._block_size = block_size
+        self._min_addr = min_addr
+        self._max_addr = max_addr
+        self._rd_perc = rd_perc
+        self._data_limit = data_limit
+        self.set_traffic()
+
+    def create_traffic(self):
```
BobbyRBruce, line 31: When you override these methods, please use my special "overrides" annotation. I find it really helps us keep a handle on things.

BobbyRBruce (changes requested):

components_library/processors/cpu_types.py
powerjg: Can we just drop this Enum? I really don't like having to specify the core types this way... Why is it needed?

BobbyRBruce: Right now the user needs to specify the CPU type of the SimpleProcessor. The alternative to this would be to have a SimpleProcessor for each CPU type: SimpleTimingProcessor, SimpleAtomicProcessor, etc, which i think would be a bit funny.

The swappableProcessor (see the parsec_disk_run.py) also makes the user select what type of core to swap to and from.

powerjg: I still don't like it. What if I wanted to have an out of order core that is 4-wide? What about an in-order core with 4 ALUs?

I think we should probably go the core route that Mahyar did and instead of specifying a *core type* give a class/factory for creating cores.

I still like the philosophy of extensible, not configurable. :)

components_library/cachehierarchies/mesi_two_level_cache_hierarchy.py
```diff
@@ -200,12 +214,17 @@ def incorporate_cache(self, motherboard: AbstractBoard) -> None:
...
+        motherboard.connect_iocache(
+            self.ruby_system.sys_port_proxy.pio_request_port
+        )
         # connect the io bus
         # TODO: This interface needs fixed. The PC should not be accessed in
         # This way
-        motherboard.pc.attachIO(
-            motherboard.get_io_bus(), [motherboard.pc.south_bridge.ide.dma]
-        )
+        # motherboard.pc.attachIO(
```
powerjg, line 228: Hack around this. I.e., `try: motherboard.pc....`

components_library/processors/abstract_generator_core.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def connect_interrupt(
+        self, interrupt_requestor: Port, interrupt_responce: Port
+    ) -> None:
+        pass
+
+    def create_idle_traffic(self):
+        yield self.dummy_generator.createIdle(0)
+
+    def setup_dummy_generator(self):
```
powerjg, line 60: If
 private functino, start iwth _

components_library/processors/py_generator_core.py
```diff
@@ -0,0 +1,49 @@
...
+
+from .abstract_generator_core import AbstractGeneratorCore
+
+
+class PyGeneratorCore(AbstractGeneratorCore):
+    def __init__(self , cpu_type: CPUTypes):
+        super(PyGeneratorCore, self).__init__(cpu_type=cpu_type)
+        self.main_generator = PyTrafficGen()
+
+    def connect_dcache(self, port: Port) -> None:
```
powerjg, line 38: Move to abstract generator

components_library/boards/test_board.py
```diff
@@ -0,0 +1,81 @@
...
+
+from .abstract_board import AbstractBoard
+from ..processors.abstract_processor import AbstractProcessor
+from ..memory.abstract_memory_system import AbstractMemorySystem
+from ..cachehierarchies.abstract_cache_hierarchy import AbstractCacheHierarchy
+
+from typing import List
+
+
+class TestBoard(AbstractBoard):
```
powerjg, line 37: COMMENTS

components_library/processors/complex_generator.py
```diff
@@ -0,0 +1,104 @@
...
+        for core in self.cores:
+            core.add_linear(
```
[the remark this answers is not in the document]
mahyarsamani: Right!

powerjg (commented):
A few small things

mahyarsamani:
Should I merge this?

BobbyRBruce:
> Should I merge this?

I'm ok with this being merged, assuming all the points have been addressed. Frankly I'd rather it submitted now in an imperfect state than let these branches diverge any more.

Let @powerjg have the final say.

powerjg (approved):
Let's do it

PR #226, opened 2021-07-31
WIP: Added Lupio-RTC device and new LupV platform

lhinman19:
Change-Id: Ibd7cb3a36d9e1a897a20c462a90e04a0d9066a1a

### DIFF
--- a/src/dev/lupio/LupioRTC.py
+++ b/src/dev/lupio/LupioRTC.py
@@ -1,0 +1,42 @@
+# Copyright (c) 2021 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects.Device import BasicPioDevice
+from m5.params import Param
+
+class LupioRTC(BasicPioDevice):
+    type = 'LupioRTC'
+    cxx_class='gem5::LupioRTC'
+    cxx_header = 'dev/lupio/lupio_rtc.hh'
+    time = Param.Time('01/01/2020', "System time to use")
+    pio_size = Param.Addr(0x1000, "PIO Size")
+
+    def generateDeviceTree(self, state):
+        node = self.generateBasicPioDeviceNode(state, "lupio-rtc",
+                                               self.pio_addr,
+                                               self.pio_size)
+        node.appendCompatible(["lupio,rtc"])
+        yield node
--- a/src/dev/lupio/SConscript
+++ b/src/dev/lupio/SConscript
@@ -1,0 +1,33 @@
+# Copyright (c) 2021 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Import('*')
+
+SimObject('LupioRTC.py')
+
+DebugFlag('LupioRTC')
+
+Source('lupio_rtc.cc')
--- a/src/dev/lupio/lupio_rtc.cc
+++ b/src/dev/lupio/lupio_rtc.cc
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2021 The Regents of the University of California
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dev/lupio/lupio_rtc.hh"
+
+#include "base/time.hh"
+#include "debug/LupioRTC.hh"
+#include "mem/packet_access.hh"
+#include "params/LupioRTC.hh"
+
+namespace gem5
+{
+
+using namespace RiscvISA;
+
+LupioRTC::LupioRTC(const Params &params) :
+    BasicPioDevice(params, params.pio_size),
+    time(params.time)
+{
+    start = mktime(&time);
+}
+
+uint8_t
+LupioRTC::lupioRTCRead(uint8_t addr)
+{
+    uint8_t r = 0;
+    uint32_t time_sec;
+    time_t total_time;
+
+    /**
+     * Get simulated time in seconds
+     */
+    time_sec = (curTick() / sim_clock::as_int::s);
+    total_time = (time_t) time_sec + start;
+
+    /**
+     * Transform into broken-down time representation
+     */
+    gmtime_r(&total_time, &time);
+
+    switch (addr) {
+        case LUPIO_RTC_SECD:
+            r = time.tm_sec;                 /* 0-60 (for leap seconds) */
+            break;
+        case LUPIO_RTC_MINT:
+            r = time.tm_min;                 /* 0-59 */
+            break;
+        case LUPIO_RTC_HOUR:
+            r = time.tm_hour;                /* 0-23 */
+            break;
+        case LUPIO_RTC_DYMO:
+            r = time.tm_mday;                /* 1-31 */
+            break;
+        case LUPIO_RTC_MNTH:
+            r = time.tm_mon + 1;             /* 1-12 */
+            break;
+        case LUPIO_RTC_YEAR:
+            r = (time.tm_year + 1900) % 100; /* 0-99 */
+            break;
+        case LUPIO_RTC_CENT:
+            r = (time.tm_year + 1900) / 100; /* 0-99 */
+            break;
+        case LUPIO_RTC_DYWK:
+            r = 1 + (time.tm_wday + 6) % 7;  /* 1-7 (Monday is 1) */
+            break;
+        case LUPIO_RTC_DYYR:
+            r = time.tm_yday + 1;            /* 1-366 (for leap years) */
+            break;
+    }
+    return r;
+}
+
+Tick
+LupioRTC::read(PacketPtr pkt)
+{
+    bool is_atomic = pkt->isAtomicOp() && pkt->cmd == MemCmd::SwapReq;
+    DPRINTF(LupioRTC,
+        "Read request - addr: %#x, size: %#x, atomic:%d\n",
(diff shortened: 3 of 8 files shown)


src/dev/lupio/LupioRTC.py
```diff
@@ -0,0 +1,19 @@
+from m5.params import *
+from m5.proxy import *
+from m5.objects.Device import BasicPioDevice
+from m5.util.fdthelper import *
```
joel-porquet, line 4: Double check that all of these imports are necessary.

src/dev/lupio/LupioRTC.py
```diff
@@ -0,0 +1,19 @@
...
+
+class LupioRTC(BasicPioDevice):
+    type = 'LupioRTC'
+    cxx_class='gem5::LupioRTC'
+    cxx_header = 'dev/lupio/lupio_rtc.hh'
+    time = Param.Time('01/01/2012',
+        "System time to use")
```
joel-porquet, line 11: Does this need to be on two lines? It doesn't seem like the line would be over 80 characters.

src/dev/lupio/LupioRTC.py
```diff
@@ -0,0 +1,19 @@
+from m5.params import *
+from m5.proxy import *
+from m5.objects.Device import BasicPioDevice
+from m5.util.fdthelper import *
+
+class LupioRTC(BasicPioDevice):
+    type = 'LupioRTC'
+    cxx_class='gem5::LupioRTC'
+    cxx_header = 'dev/lupio/lupio_rtc.hh'
+    time = Param.Time('01/01/2012',
```
joel-porquet, line 10: Is there a reason behind this date? Or is it just a copy from another file? What about the typical 01/01/1970 otherwise?

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
...
+#include <string>
+
+#include "base/bitfield.hh"
+#include "base/time.hh"
+#include "base/trace.hh"
+#include "debug/LupioRTC.hh"
+#include "mem/packet.hh"
+#include "mem/packet_access.hh"
+#include "params/LupioRTC.hh"
+#include "sim/system.hh"
```
joel-porquet, line 13: Double check these includes as well.

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
...
+    system(params.system),
+    time(params.time)
+{
+    struct timeval tv;
+    struct tm time_bd;
+
+    gettimeofday(&tv, NULL);
+    gmtime_r(&tv.tv_sec, &time_bd);
+
+    DPRINTFN("Real-time clock set to %s", asctime(&time_bd));
```
joel-porquet, line 31: I'm not sure what this constructor is supposed to do. Both variables are local, so they will disappear when the function returns.

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
...
+        case LUPIO_RTC_CENT:
+            r = (time_bd.tm_year + 1900) / 100;
+            break;
+        case LUPIO_RTC_DYWK:
+            r = 1 + (time_bd.tm_wday + 6) % 7;
+            break;
+        case LUPIO_RTC_DYYR:
+            r = time_bd.tm_yday + 1;
+            break;
+    }
```
joel-porquet, line 76: Make sure to keep the same comments as the QEMU model to explain the format of each field. Otherwise, it looks a bit random.

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
...
+
+    pkt->setData(&time_val);
+    pkt->makeAtomicResponse();
+    return pioDelay;
+}
+
+Tick
+LupioRTC::write(PacketPtr pkt)
+{
+    return pioDelay;
```
joel-porquet, line 100: Is it the way to have the write operation be "unsupported"?

src/dev/lupio/lupio_rtc.hh
```diff
@@ -0,0 +1,59 @@
...
+    static const int LUPIO_RTC_SECD = 0x00;
+    static const int LUPIO_RTC_MINT = 0x01;
+    static const int LUPIO_RTC_HOUR = 0x02;
+    static const int LUPIO_RTC_DYMO = 0x03;
+    static const int LUPIO_RTC_MNTH = 0x04;
+    static const int LUPIO_RTC_YEAR = 0x05;
+    static const int LUPIO_RTC_CENT = 0x06;
+    static const int LUPIO_RTC_DYWK = 0x07;
+    static const int LUPIO_RTC_DYYR = 0x08;
+    static const int LUPIO_RTC_MAX = 0x09;
```
joel-porquet, line 36: You could probably use an enum here.

src/dev/lupio/lupio_rtc.hh
```diff
@@ -0,0 +1,59 @@
...
+
+  public:
+    PARAMS(LupioRTC);
+    LupioRTC(const Params &params);
+    virtual ~LupioRTC();
+
+    /**
+    * Function to return current time and date using system's wall-clock time
+    */
+
```
joel-porquet, line 46: Remove blank line between comment and function protype? (check with coding convention in gem5)

src/dev/lupio/lupio_rtc.hh
```diff
@@ -0,0 +1,59 @@
...
+    uint8_t lupioRTCRead(const uint8_t addr);
+
+    /**
+    * Implement BasicPioDevice virtual functions
+    */
+
```
joel-porquet, line 52: Same here, can probably remove the blank line.

src/dev/riscv/LupV.py
```diff
@@ -0,0 +1,182 @@
+from m5.objects.Platform import Platform
+from m5.objects.PMAChecker import PMAChecker
+from m5.objects.Clint import Clint
+from m5.objects.LupioRTC import LupioRTC
+from m5.objects.Plic import Plic
+from m5.objects.RTC import RiscvRTC
+from m5.objects.Uart import Uart8250
+from m5.objects.Terminal import Terminal
```
joel-porquet, line 8: Order imports alphabetically.

src/dev/riscv/LupV.py
```diff
@@ -0,0 +1,182 @@
...
+from m5.params import *
+from m5.proxy import *
+from m5.util.fdthelper import *
```
joel-porquet, line 11: Again, double check that each import is necessary.

joel-porquet (changes requested):
Great first PR! :)

I left a few comments for the code, and as mentioned in the chat, it'd be great to split the commit into two.

src/dev/lupio/LupioRTC.py
powerjg: How about 1/1/2020 for gem5-20 (kinda)?

We probably don't want to use 1/1/1970 in case some installed program cares about the time. It may be weird to see the time be so old.

src/dev/lupio/LupioRTC.py
```diff
@@ -0,0 +1,19 @@
+from m5.params import *
```
powerjg, line 1: This file needs a license text. Copyright should be UC Regents. See src/arch/riscv/pmp.cc for an example.

src/dev/lupio/LupioRTC.py
powerjg: To add, avoid `import *` if you can.

src/dev/lupio/SConscript
```diff
@@ -0,0 +1,9 @@
+# -*- mode:python -*-
```
powerjg, line 1: License. Also, remove the mode:python

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
+#include "dev/lupio/lupio_rtc.hh"
```
powerjg, line 1: License :)

src/dev/lupio/lupio_rtc.cc
powerjg: Also on this line. Instead of `DPRINTFN`, it would be better to use `DPRINTF(LupioRTC, ...`

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
...
+    struct timeval tv;
+    struct tm time_bd;
+
+    gettimeofday(&tv, NULL);
+    gmtime_r(&tv.tv_sec, &time_bd);
+
+    DPRINTFN("Real-time clock set to %s", asctime(&time_bd));
+}
+
+LupioRTC::~LupioRTC()
```
powerjg, line 34: Don't need this function if it's empty. If it's pure virtual (which I don't think it is...) I *think* you can use `= default` in the header.

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
...
+    DPRINTF(LupioRTC,
+        "Read request - addr: %#x, size: %#x, atomic:%d\n",
+        pkt->getAddr(), pkt->getSize(), is_atomic);
+
+    Addr daddr = pkt->getAddr() - pioAddr;
+    uint8_t time_val = lupioRTCRead(daddr);
+    DPRINTF(LupioRTC, "time: %d\n", time_val);
+
+    pkt->setData(&time_val);
+    pkt->makeAtomicResponse();
```
powerjg, line 93: Have you tested this in timing mode? I think this won't work. I believe you'll need to say:

```suggestions
    if(is_atomic) {
        pkt->makeAtomicResponse();
    } else {
        pkt->makeResponse();
    }
```

src/dev/lupio/lupio_rtc.cc
powerjg: There's a couple of options here:

- Ignore writes like you do now.
- Print a warning `warn(...)` if you get a write to this device.
- Raise an error if you get a write: `panic("Unexpected write to the LupioRTC device!")`

I think the last is probably the right thing to do since this is for *teaching*. If a student writes a wrong driver this would catch it.

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,103 @@
...
+    return pioDelay;
+}
+
+Tick
+LupioRTC::write(PacketPtr pkt)
+{
+    return pioDelay;
+}
+
+}
```
powerjg, line 103:
```suggestion
} // namespace gem5
```

It makes it easier to read :)

src/dev/lupio/lupio_rtc.hh
```diff
@@ -0,0 +1,59 @@
...
+#include "params/LupioRTC.hh"
+#include "sim/core.hh"
+#include "sim/system.hh"
+
+namespace gem5
+{
+
+using namespace RiscvISA;
+
+/** Real-Time Clock (LUPIO_RTC) */
```
powerjg, line 17: Adding a more detailed doxygen comment would be great.

src/dev/lupio/lupio_rtc.hh
```diff
@@ -0,0 +1,59 @@
...
+  public:
+    PARAMS(LupioRTC);
+    LupioRTC(const Params &params);
+    virtual ~LupioRTC();
+
+    /**
+    * Function to return current time and date using system's wall-clock time
+    */
+
+    uint8_t lupioRTCRead(const uint8_t addr);
```
powerjg, line 47: This should probably be private. I don't think you would expect another object to call this function.

src/dev/riscv/LupV.py
```diff
@@ -0,0 +1,182 @@
+from m5.objects.Platform import Platform
```
powerjg, line 1: License

src/dev/lupio/lupio_rtc.hh
```diff
@@ -0,0 +1,59 @@
+#ifndef __DEV_LUPIO_LUPIO_RTC_HH__
```
powerjg, line 1: License :)

src/dev/riscv/LupV.py
```diff
@@ -0,0 +1,182 @@
...
+from m5.objects.Plic import Plic
+from m5.objects.RTC import RiscvRTC
+from m5.objects.Uart import Uart8250
+from m5.objects.Terminal import Terminal
+from m5.params import *
+from m5.proxy import *
+from m5.util.fdthelper import *
+
+class LupV(Platform):
+    """LupV Platform
```
powerjg, line 14: Maybe a comment here saying that you copied this from HiFive.py and that you'll be updating it over a number of commits.

src/dev/riscv/lupv.cc
```diff
@@ -0,0 +1,62 @@
+#include "dev/riscv/lupv.hh"
```
powerjg, line 1: License

src/dev/riscv/lupv.hh
```diff
@@ -0,0 +1,45 @@
+#ifndef __DEV_RISCV_LUPV_HH__
```
powerjg, line 1: License

powerjg (commented):
There's probably a bit of a race condition here. Sorry!

One high-level thing... once we are able to drop the PLIC/CLINT we may want to revisit the need for the "platform" at all. I don't fully understand what that SimObject does...

src/dev/lupio/LupioRTC.py
joel-porquet: Funnily enough, I believe that raspberryPi don't have an RTC, so if they boot for the first time without network access to set up NTP, they actually show Jan 1st 1970!

powerjg:
For the commit messages, I would use "dev,riscv: ..." for the lupv platform and "dev:" for the RTC.

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,128 @@
...
+        pkt->makeResponse();
+    }
+
+    return pioDelay;
+}
+
+Tick
+LupioRTC::write(PacketPtr pkt)
+{
+    panic("Unexpected write to the LupioRTC device!");
```
joel-porquet, line 124: Are you sure that panicking is the right way to go? Isn't there a more gentle way to go about it: like print a warning and/or return some error to the CPU as part of the simulation?

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,128 @@
...
+
+uint8_t
+LupioRTC::lupioRTCRead(uint8_t addr)
+{
+    uint8_t r = 0;
+
+    /**
+     * Get real time in seconds
+     */
+    gettimeofday(&tv, NULL);
```
joel-porquet, line 59: I don't think your code compiles, does it? `tv` hasn't been declared.

src/dev/lupio/lupio_rtc.cc
```diff
@@ -0,0 +1,128 @@
...
+using namespace RiscvISA;
+
+LupioRTC::LupioRTC(const Params &params) :
+    BasicPioDevice(params, params.pio_size),
+    time(params.time)
+{
+    gettimeofday(&tv, NULL);
+    gmtime_r(&tv.tv_sec, &time);
+
+    DPRINTF(LupioRTC, "Real-time clock set to %s", asctime(&time));
```
joel-porquet, line 48: `tv` and `time` haven't been declared, and I'm still unclear on what this constructor is trying to do. Can you explain?

src/dev/lupio/lupio_rtc.hh
```diff
@@ -0,0 +1,93 @@
...
+        LUPIO_RTC_DYMO,
+        LUPIO_RTC_MNTH,
+        LUPIO_RTC_YEAR,
+        LUPIO_RTC_CENT,
+        LUPIO_RTC_DYWK,
+        LUPIO_RTC_DYYR,
+
+        /**
+         * Max offset
+         */
```
joel-porquet, line 71: For small comments like that, wouldn't `/* ... */` or `//` be enough? (confirm with coding conventions in gem5)

src/dev/lupio/LupioRTC.py
```diff
@@ -0,0 +1,43 @@
...
+
+from m5.objects.Device import BasicPioDevice
+from m5.params import Param
+from m5.util.fdthelper import *
+
+class LupioRTC(BasicPioDevice):
+    type = 'LupioRTC'
+    cxx_class='gem5::LupioRTC'
+    cxx_header = 'dev/lupio/lupio_rtc.hh'
+    time = Param.Time('01/01/2020', "System time to use")
```
joel-porquet, line 35: This parameter doesn't seem to be actually used, since the RTC will recompute the time every time an access is made. Is there a reason to have it?

joel-porquet (changes requested):

src/dev/lupio/lupio_rtc.hh
powerjg: I think doxygen supports `///`

src/dev/lupio/LupioRTC.py
powerjg: Let's chat about this tomorrow. I think we need to make a group decision on how to model the "time" in the RTC.

src/dev/lupio/lupio_rtc.cc
powerjg: I think this is the right way to do it. There's two ways to look at it.
1. If the simulator has a bug and delivers a write to this device, we want to panic, it's a bug.
2. If a student is writing a device driver and does something wrong, rather than having to write good error handling code and checking the error correctly, I think we want to "fail fast" and have the simulator terminate. This is, in some sense, causing a "puff of smoke" out of the hardware.

src/dev/lupio/lupio_rtc.cc
mkjost0: Currently, we have both `tv` and `time` defined in `lupio_rtc.hh`, because we use them in both the constructor and in LupioRTCRead.  In the constructor, we just wanted to grab the current time so we could use a debug flag to look at the time being saved while we were testing, since it was the only function that we were sure was being called while we still weren't finished.  However, we could now instead move this print statement to LupioRTCRead, make `tv` and `time` local variables to that function, and empty the constructor if that makes more sense to do?

joel-porquet: Unless you really initialize the time in the constructor (in which case the device model needs to be somewhat rewritten to take that in account --but in that case, your variable should be part of the class itself), then you could just empty the constructor and the printf could be move to the read function (or even dropped altogether).

src/dev/lupio/lupio_rtc.cc
powerjg: A bus error would be a property of the bus protocol, right? What is the bus protocol for the RTC? It's just wires. There's no way to signal an error without having some kind of higher-level protocol in place. There's no default protocol in gem5 (or in any hardware). 

We can certainly just ignore writes, but that could mask bugs. I still think panic-ing (as defined by gem5) is the right thing to do. However, if you have an example where this device is written and it's not either a hardware or software bug, I can change my mind :)

joel-porquet: Fine by me if it's the right way to go :+1:

src/dev/riscv/LupV.py
lhinman19: I double checked the imports and am now wondering if it is ok to have the `import *` for `params` and `fdthelper`?

powerjg: Seems fine to me :)

powerjg (commented):
Looks great to me!

The final thing is the commit messages. The tags should probably be "dev:" for the one that adds the lupio RTC device and "dev,arch-riscv:" for the one that adds the lupv platform.

Then, you need to add enough info to the commit message so someone can review the commits. For instance, you need to include a link to the details of the LupIO devices (the specifications?) and some info about how you implemented it

joel-porquet (changes requested):

src/dev/lupio/SConscript
```diff
@@ -0,0 +1,33 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Import('*')
+
+SimObject('LupioRTC.py')
+
+DebugFlag('LupioRTC')
+
+Source('lupio_rtc.cc')
```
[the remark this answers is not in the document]
powerjg: That's not the gem5 style :). The file names match the class names 90 percent of the time.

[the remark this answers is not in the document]
joel-porquet: Fine by me if it's the convention in gem5. But it *will* be a pain to navigate through files :P

src/dev/lupio/LupioRTC.py
powerjg: No, because that wouldn't be deterministic. It's important that every time you run a benchmark in gem5 that you get the same result.

If the user wants to use now(), they can do that from the config script easily.

joel-porquet: Hmm, I forgot about the deterministic part. Ok, fine then.

joel-porquet:
Looks all good to me! I'll let you all handle the switch from `develop` to `lupio` for the branch to be merged in.

Happy Friday :smiley:

PR #227, opened 2021-08-12
Components/jason/riscv fs

powerjg:


components_library/boards/riscv_board.py
```diff
@@ -0,0 +1,280 @@
...
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from configs.common.Caches import IOCache
```
BobbyRBruce, line 27: I really don't like relying on anything in `configs` inside the component library. While stuff in configs can use the components library, I don't think otherway around should be permitted. If this is needed, could we migrate it?

components_library/boards/riscv_board.py
```diff
@@ -0,0 +1,280 @@
...
+import os
+from typing import Optional
+
+from components_library.utils.override import overrides
```
BobbyRBruce, line 31: It's not a big deal, but in general I've been using relative paths as much as possible within the library.

components_library/boards/riscv_board.py
```diff
@@ -0,0 +1,280 @@
...
+                else:
+                    root.append(node)
+
+        fdt = Fdt()
+        fdt.add_rootnode(root)
+        fdt.writeDtsFile(os.path.join(outdir, "device.dts"))
+        fdt.writeDtbFile(os.path.join(outdir, "device.dtb"))
+
+    # TODO: Finish implementing this.
+    def generateDeviceTree(self, state):
```
BobbyRBruce, line 263: This function is never called, right?

components_library/cachehierarchies/classic/no_cache.py
```diff
@@ -113,6 +113,8 @@ def incorporate_cache(self, board: AbstractBoard) -> None:
                 int_req_port = self.membus.mem_side_ports
                 int_resp_port = self.membus.cpu_side_ports
                 core.connect_interrupt(int_req_port, int_resp_port)
+            else:
+                core.connect_interrupt()
```
BobbyRBruce, line 117: Just to clarify, this holds true for all non-X86 targets?

components_library/notes.md
```diff
@@ -0,0 +1,62 @@
...
+  - Make is so that we don't have to use the variable name "disk" for the platform device
+- Check if it works with no_cache. It should as the fs_linux.py file seems to.
+
+- MMU problems
+  - Need to have MMU caches when there are caches
+  - Or, if there are no caches, then need to have another bus to connect MMUs
+
+## General components improvements
+
+- Add a "CompatibilityError" instead of "EnvironmentError"
```
BobbyRBruce, line 28: Yes, i think someone should sit down and create some bespoke gem5 Errors. I've not been thinking very hard about this and generally throwing rather generic exceptions. There's a Jira ticket for this.

components_library/notes.md
```diff
@@ -0,0 +1,62 @@
...
+- make `is_ruby` a member of the cache hierarchy, not an "environment" variable.
```
BobbyRBruce, line 29: It is isn't it? It's an abtract function in AbstractCacheHierarchy.

components_library/notes.md
```diff
@@ -0,0 +1,62 @@
...
+- Add a helper for disk images?
+- Limitations of SimpleBoard doesn't seem true anymore
```
BobbyRBruce, line 31: If we can support full system on the simple board, that'd be great!

components_library/notes.md
```diff
@@ -0,0 +1,62 @@
...
+- When can I use the `get_processor` and `get_cores` functions?
```
BobbyRBruce, line 32: Whenever you want. As soon as these functions are available they are usable. The processor exists within a board as soon as you construct the board. The processor contains the cores as soon as its constructed. `incorporate_processor` is simply connecting things to the board correctly.

components_library/notes.md
```diff
@@ -0,0 +1,62 @@
...
+- In simple_core can't import all CPU types (e.g., KVM CPU) as they may not be built
+- membus should be a class member, and we shouldn't set the badaddr responder every time _get_default_membus is called
+- I don't love the way the createInterrupts() functions are called. This should probably be either in the `__init__` function or in a different function than `connect_interrupts`. I think the board should be responsible for this, not the caches.
+
+## Questions about components
+
+- What's the point of SimpleBoard and why does X86Board inherit it?
+  - Has some useful functions... document this
+- Where do I get the memory size? Is this a parameter memory that the system should grab, or is it a parameter that system should set on the memory?
```
BobbyRBruce, line 41: board.get_memory().get_memory_ranges()

components_library/notes.md
```diff
@@ -0,0 +1,62 @@
...
+
+- Should we use the bare metal workload or `RiscvLinux`?
+  - The difference is the dtb file baremetal assumes the dtb file is compiled into the bootloader. The `RiscvLinux` workload can load a the dts file from disk.
+- How much of HiFive platform could be moved out into the components?
+- How does virtio disk image work?
+- The PMA checker shouldn't have an empty default in the RiscvMMU. This is a recipe for disaster
+
+## Things that could be useful
+
+A `require` function which, for instance, requires that ISA is RISCV.
```
BobbyRBruce, line 55: Yes, I think a more clever way to delcare run-time requirements would be useful and remove a lot of boilerplate. For example, when I write a config script I'm having to put a lot of if-statements at the top making sure everything is ok, otherwise return an error (e.g., "You need to have gem5 compiled to X86").

components_library/notes.md
```diff
@@ -0,0 +1,62 @@
+# Notes setting up RISC-V FS in components
```
BobbyRBruce, line 1: Lots of good stuff here, I'll add some of the more interesting ones as Jira issues this afternoon.

components_library/processors/simple_core.py
```diff
@@ -95,3 +97,7 @@ def connect_interrupt(
             self.core.interrupts[0].pio = interrupt_requestor
             self.core.interrupts[0].int_requestor = interrupt_responce
             self.core.interrupts[0].int_responder = interrupt_requestor
```
BobbyRBruce, line 99: Does gem5 work ok if you set these to None?

BobbyRBruce (commented):
Thanks Jason. I think this is really good. It gives me more confidence in our current design a you've not needed to alter too much to get this working. I've left some comments but nothing major. I enjoyed reading your notes :)

components_library/boards/riscv_board.py
powerjg: Whoa, this was a big mistake on my part! I blame VS code for auto importing it :)

components_library/boards/riscv_board.py
powerjg: Correct. Hence the TODO

components_library/cachehierarchies/classic/no_cache.py
powerjg: It's definitely true that all targets must have `BaseCPU.createInterrupts()` called on it. Whether this is the right place to do it or not is up for debate :)

components_library/notes.md
powerjg: Yeah, I found that after making this note. That's not how the x86 board does it though.

components_library/notes.md
powerjg: I see what you're saying now. The `SimpleBoard` only supports SE mode by itself.

What I am saying is that things that derive from `SimpleBoard` support FS. I think the documentation is a bit confusing. I am on the fence as to whether we should have a separate SE vs FS base board.

components_library/notes.md
powerjg: We should document this :)

components_library/processors/simple_core.py
powerjg: Hmm, probably not. They are required when x86 and not required otherwise. Maybe instead the condition should be `if interrupt_requestor or interrupt_responder` or something like that instead of checking for x86 multiple places.

components_library/memory/abstract_memory_system.py
```diff
@@ -41,17 +41,33 @@ def __init__(self) -> None:
...
+        """Get the ports to connect this memory system to the cache"""
         raise NotImplementedError
 
     @abstractmethod
     def get_memory_controllers(self) -> List[MemCtrl]:
+        """Get all of the memory controllers in this memory system"""
         raise NotImplementedError
 
     @abstractmethod
-    def get_memory_ranges(self) -> List[AddrRange]:
```
BobbyRBruce, line 55: Surely we still want to be able to get the memory ranges?

components_library/boards/x86_board.py
```diff
@@ -94,10 +95,17 @@ def __init__(
...
+            )
+        data_range = AddrRange(memory.get_size())
+        memory.set_memory_range([data_range])
+
         # Add the address range for the IO
-        # TODO: This should definitely NOT be hardcoded to 3GB
         self.mem_ranges = [
-            AddrRange(Addr("3GB")),  # All data
+            data_range,  # All data
             AddrRange(0xC0000000, size=0x100000),  # For I/0
```
BobbyRBruce, line 109: There's nothing stopping these ranges overlapping right now, right?

components_library/memory/abstract_memory_system.py
```diff
@@ -41,17 +41,33 @@ def __init__(self) -> None:
...
-        """All of the memory ranges that this memory system responds to."""
+    def get_size(self) -> int:
+        """Returns the total size of the memory system"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def set_memory_range(self, ranges: List[AddrRange]) -> None:
```
BobbyRBruce, line 64: I see why you've done it this way, but I deeply dislike having to set the memory ranges in this way. At the core of my design is that all the board-component interaction should happen with the "incorporate_x" functions. Do you think there's a way you could migrate that there? Even then I'm not super happy with that as it's something someone might forget about :/

A more complex design to enforce this could be:

```
def incorporate_memory(...):
   _incorporate_memory(...)
   _set_memory_range(...)

@abstractmemory
def _incorporate_memory(...):
   pass

@abstractmethod
def _set_memory_range(...)
   pass
```

in the AbstractMemorySystem.

Just a thought, willing to discuss this.

components_library/processors/simple_core.py
BobbyRBruce: Don't forget to implement this :)

components_library/memory/abstract_memory_system.py
powerjg: Why? I would argue that should be hidden since the board should be in charge of this.

components_library/boards/x86_board.py
powerjg: There's a check to make sure data_range is < 3GB. That's the best we can do right now.

components_library/memory/abstract_memory_system.py
powerjg: Done!

components_library/processors/simple_core.py
powerjg: Do I need to do something more right now? Currently, it's checking to see if it's x86. We could probably clean that up, but it's not currently in my scope. I have a few other open things in this PR that I'll try to come back to once I've pushed this relation chain upstream.

components_library/boards/test_board.py
```diff
@@ -61,12 +68,12 @@ def __init__(
             clock=clk_freq, voltage_domain=VoltageDomain()
         )
 
-        self.mem_ranges = memory.get_memory_ranges()
-
     def connect_system_port(self, port: Port) -> None:
         self.system_port = port
 
     def connect_things(self) -> None:
```
BobbyRBruce, line 74: Is there a good reason to not move this implementation to the `AbstractBoard`? This seems like pretty default behavior (I know X86 will need overridden regardless).

components_library/boards/abstract_board.py
```diff
@@ -167,6 +165,26 @@ def set_mem_mode(self, mem_mode: MemMode) -> None:
...
+        the size of memory and memory's memory range to be the same as the
+        board. Full system implementations will likely need something more
+        complicated.
+        """
+        memory = self.get_memory()
+
+        # The simple board just has one memory range that is the size of the
+        # memory.
+        self.mem_ranges = [AddrRange(memory.get_size())]
+        memory.set_memory_range(self.mem_ranges)
```
BobbyRBruce, line 186: I think, given your documentation here, you meant this to go in the Simple board, and this function to be abstract? That's how i'd do it.

components_library/boards/x86_board.py
```diff
@@ -267,7 +268,25 @@ def assignISAInt(irq, apicPin):
...
-        super().connect_things()
+        # This board is a bit particular about the order that things are
+        # connected together.
+
+        # Before incorporating the memory or creating the I/O devices figure
+        # out the memory ranges.
+        self.setup_memory_ranges()
+
+        # Set up all of the I/O before we incorporate anything else.
+        self._setup_io_devices()
```
BobbyRBruce, line 279: This is quite annoying as it kind of breaks out design a little bit. I think a better way to do this would be to call `setup_memory_ranges()` in the `AbstractBoard` constructor. I know this is a _little bit_ ugly, but it allows us to develop new boards with the assumption that the memory ranges are already setup after construction of the base class. In this case, the `_setup_io_devices()` can be called in the constructor of the x86_board, and `connect_things()` can be kept simple.

components_library/memory/abstract_memory_system.py
BobbyRBruce: Fair point. I was battling over what should be in charge of memory ranges, but yes, the board is the place for it to be.

components_library/processors/simple_core.py
BobbyRBruce: Ok. It's probably fine right now.

BobbyRBruce (commented):
Thanks Jason. I think this is really starting to come together and is strengthening our design. I've left some comments about moving stuff around a little.

components_library/boards/test_board.py
powerjg: Yeah, you're right. That was a step I forgot to take.

components_library/boards/abstract_board.py
powerjg: Wait... why would this go in simple board and connect_things go in abstract? They should both be in one or the other.

components_library/boards/x86_board.py
powerjg: I thought about that. But then you have to make sure to call super().__init__ at the *end* of the init function.

powerjg:
Closing this PR since it's on gerrit now.

PR #228, opened 2021-08-13
WIP: Added new Lupio-RNG device

mkjost0:
Change-Id: If9800d1e05b4c28299c2a7ade51a21762d52a40e

### DIFF
--- a/src/dev/lupio/LupioRNG.py
+++ b/src/dev/lupio/LupioRNG.py
@@ -24,13 +24,35 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-Import('*')
+from m5.objects.Device import BasicPioDevice
+from m5.params import Param
+from m5.proxy import Parent
+from m5.defines import buildEnv
+from m5.util.fdthelper import FdtPropertyWords
 
-SimObject('LupioRNG.py')
-SimObject('LupioRTC.py')
+class LupioRNG(BasicPioDevice):
 
-DebugFlag('LupioRNG')
-DebugFlag('LupioRTC')
+    type = 'LupioRNG'
+    cxx_class='gem5::LupioRNG'
+    cxx_header = 'dev/lupio/lupio_rng.hh'
+    pio_size = Param.Addr(0x1000, "PIO Size")
+    seed = Param.Int(0, "Initial seed for the random number generator")
+    if buildEnv['TARGET_ISA'] == 'riscv':
+        plic = Param.Plic(Parent.any, "PLIC")
+        lupio_rng_int_id = Param.Int(3, "Interrupt ID for the RNG")
 
-Source('lupio_rng.cc')
-Source('lupio_rtc.cc')
+    def generateDeviceTree(self, state):
+
+        if buildEnv['TARGET_ISA'] != 'riscv':
+            return
+
+        node = self.generateBasicPioDeviceNode(state, "lupio-rng",
+                                               self.pio_addr,
+                                               self.pio_size)
+        plic = self.plic
+        node.append(
+                FdtPropertyWords("interrupts", [self.lupio_rng_int_id]))
+        node.append(
+                FdtPropertyWords("interrupt-parent", state.phandle(plic)))
+        node.appendCompatible(["lupio,rng"])
+        yield node
--- a/src/dev/lupio/lupio_rng.cc
+++ b/src/dev/lupio/lupio_rng.cc
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2021 The Regents of the University of California
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dev/lupio/lupio_rng.hh"
+
+#include <random>
+
+#include "debug/LupioRNG.hh"
+#include "mem/packet_access.hh"
+#include "params/LupioRNG.hh"
+
+/* CTRL fields */
+#define LUPIO_RNG_IRQE  0x1
+/* STAT fields */
+#define LUPIO_RNG_BUSY  0x1
+
+namespace gem5
+{
+
+LupioRNG::LupioRNG(const Params &params) :
+    BasicPioDevice(params, params.pio_size),
+    user_seed(params.seed)
+{
+    mt.seed(user_seed);
+    DPRINTF(LupioRNG, "LupioRNG initalized with seed: %d\n", user_seed);
+}
+
+uint64_t
+LupioRNG::lupioRNGRead(uint8_t addr)
+{
+    uint32_t r = 0;
+    std::uniform_int_distribution<uint32_t> distrib(0);
+
+    switch (addr >> 2) {
+        // returns a random number
+        case LUPIO_RNG_RAND:
+            r = distrib(mt);
+            DPRINTF(LupioRNG, "Random number: %#x\n", r );
+            break;
+        // returns the current seed being used
+        case LUPIO_RNG_SEED:
+            r = user_seed;
+            DPRINTF(LupioRNG, "RNG_SEED Call: %d\n", r);
+            break;
+        // returns status of the device
+        case LUPIO_RNG_STAT:
+            r = 0;  /* Always ready */
+            DPRINTF(LupioRNG, "RNG_STAT Call: %d\n", r);
+            break;
+        default:
+            r = -1;
+            panic("Unexpected read to the LupioRTC device at address %d!",
+                    addr);
+            break;
+    }
+    return r;
+}
+
+void
+LupioRNG::lupioRNGWrite(uint8_t addr, uint64_t val64)
+{
+    switch (addr >> 2) {
+        // allows the user to configure the seed value
+        case LUPIO_RNG_SEED:
+            user_seed = val64;
+            mt.seed(user_seed);
+            DPRINTF(LupioRNG, "RNG_SEED_WRITE: %d\n", user_seed);
+            break;
+        // configures the device using interrupts
+        case LUPIO_RNG_CTRL:
+            interrupt_enable = !!(val64 & LUPIO_RNG_IRQE);
+            DPRINTF(LupioRNG, "RNG_CTRL Call: %d\n", interrupt_enable);
+            break;
+        default:
+            panic("Unexpected write to the LupioRTC device at address %d!",
+                    addr);
+            break;
+    }
+}
+
+Tick
+LupioRNG::read(PacketPtr pkt)
+{
+    Addr rng_addr = pkt->getAddr() - pioAddr;
+
+    DPRINTF(LupioRNG,
+        "Read request - addr: %#x, size: %#x\n", rng_addr, pkt->getSize());
+
+    uint64_t rand_read = lupioRNGRead(rng_addr);
+    DPRINTF(LupioRNG, "Packet Read: %#x\n", rand_read);
+    pkt->setUintX(rand_read, byteOrder);
+    pkt->makeResponse();
+
+    return pioDelay;
+}
+
+Tick
+LupioRNG::write(PacketPtr pkt)
+{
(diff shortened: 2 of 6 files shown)


src/dev/lupio/LupioRNG.py
```diff
@@ -0,0 +1,53 @@
...
+from m5.params import Param
+from m5.proxy import Parent
+from m5.util.fdthelper import *
+
+
+class LupioRNG(BasicPioDevice):
+    type = 'LupioRNG'
+    cxx_class='gem5::LupioRNG'
+    cxx_header = 'dev/lupio/lupio_rng.hh'
+    pio_size = Param.Addr(0x1000, "PIO Size")
```
powerjg, line 37: Do you really need 4 KB? (Maybe so since that's the page size?)

Would 16 B work, I wonder? (e.g., `0x10`). It might be worth a try.

src/dev/lupio/LupioRNG.py
```diff
@@ -0,0 +1,53 @@
...
+    byte_order = Param.ByteOrder("big", "Device byte order")
+    platform = Param.Platform(Parent.any, "Platform this device is part of.")
+    seed = Param.Int(0, "Seed")
```
powerjg, line 40: Maybe "initial seed for the random number generator"?

src/dev/lupio/LupioRNG.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def generateDeviceTree(self, state):
+        node = self.generateBasicPioDeviceNode(state, "lupio_rng",
+                                               self.pio_addr,
+                                               self.pio_size)
+        platform = self.platform.unproxy(self)
+        plic = platform.plic
+        node.append(
+                FdtPropertyWords("interrupts", [platform.lupio_rng_int_id]))
```
powerjg, line 49: Rather than unproxying the platform, why not just pass the int_id in as a parameter? Same with the plic.

That said, some of this device tree generation stuff requires circular dependencies it seems. So, I could be wrong here.

src/dev/lupio/lupio_rng.cc
```diff
@@ -0,0 +1,138 @@
...
+{
+    mt.seed(user_seed);
+    DPRINTF(LupioRNG, "LupioRNG initalized with seed: %d\n", user_seed);
+}
+
+uint64_t
+LupioRNG::lupioRNGRead(uint8_t addr)
+{
+    uint32_t r = 0;
+    switch (addr >> 2) {
```
powerjg, line 61: This formatting is a bit weird. That said, I don't think we've said how it should be :). Here's the Google style guide: https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements

Also, I would suggest using curly braces everywhere or nowhere in the case statements.

src/dev/lupio/lupio_rng.cc
```diff
@@ -0,0 +1,138 @@
...
+        case LUPIO_RNG_SEED:
+            user_seed = val64;
+            mt.seed(user_seed);
+            DPRINTF(LupioRNG, "RNG_SEED_WRITE: %d\n", user_seed);
+            break;
+        // configures the device using interrupts
+        case LUPIO_RNG_CTRL:
+            interrupt_enable = !!(val64 & LUPIO_RNG_IRQE);
+            DPRINTF(LupioRNG, "RNG_CTRL Call: %d\n", interrupt_enable);
+            break;
```
powerjg, line 102: There's no default. Is that on purpose?

powerjg (commented):
Can you update this PR so it's merging into lupio/main so we don't see the changes from the RTC here as well?

src/dev/lupio/LupioRNG.py
mkjost0: For this, I was following the setup of all the LupIO devices in QEMU, since in the end it seemed that we wanted each device to have the same size.  However, I can change it to be smaller if we want it to be.

powerjg: TBH, I don't know what the constraints are on this. Although, if we do expose more addresses than the hardware actually has, we should be sure to check the incoming address in the hardware is a valid addr (which we do for reads, but not writes right now)

joel-porquet: FYI, the RTC was merged with a 4KiB range as well. So we could do the same here, but as Jason says, invalid addresses should be manually caught.

src/dev/lupio/LupioRNG.py
```diff
@@ -0,0 +1,53 @@
...
+    type = 'LupioRNG'
+    cxx_class='gem5::LupioRNG'
+    cxx_header = 'dev/lupio/lupio_rng.hh'
+    pio_size = Param.Addr(0x1000, "PIO Size")
+    byte_order = Param.ByteOrder("big", "Device byte order")
+    platform = Param.Platform(Parent.any, "Platform this device is part of.")
+    seed = Param.Int(0, "Seed")
+
+    def generateDeviceTree(self, state):
+        node = self.generateBasicPioDeviceNode(state, "lupio_rng",
```
joel-porquet, line 43: Replace _ with - for node name

src/dev/lupio/SConscript
```diff
@@ -0,0 +1,36 @@
...
+Import('*')
+
+SimObject('LupioRNG.py')
+SimObject('LupioRTC.py')
+
+DebugFlag('LupioRNG')
+DebugFlag('LupioRTC')
+
+Source('lupio_rng.cc')
+Source('lupio_rtc.cc')
```
joel-porquet, line 36: Oh yeah, if you branch out of `lupio/main`, this should not happen. This file already exists there, it shouldn't be created entirely.

src/dev/lupio/lupio_rng.cc
```diff
@@ -0,0 +1,138 @@
...
+        // returns status of the device
+        case LUPIO_RNG_STAT:
+            r = 0;  /* Always ready */
+            DPRINTF(LupioRNG, "RNG_STAT Call: %d\n", r);
+            break;
+        default:
+            r = -1;
+            DPRINTF(LupioRNG, "Faulty Address: %d\n", addr);
+            break;
```
joel-porquet, line 83: For the RTC, the decision was to panic upon invalid accesses. Should it be the same here?

joel-porquet (changes requested):
It seems like the target branch for this PR is `develop` when it should be `lupio/main`. Many of the files in this PR are recreated from scratch, when there should only be few modified lines.

Could you also split your commit in two: the first would introduce the new component in src/dev/lupio, the second would use this component in the LupV platform. This way each commit focuses on one logical change only.

Thanks!

joel-porquet:
Does `lupio/main` really need to be rebased onto `develop`?
- If no, then undo the rebase and just push the two new commits on top of the former `lupio/main` branch.
- If yes, then I would suggest pushing the updated `lupio/main` branch directly without a PR, and then submitting a PR that only contains the two new commits.

Bottom line, this PR is difficult to review because it contains a rebase in addition to the actual commits.

src/dev/lupio/LupioRNG.py
```diff
@@ -0,0 +1,53 @@
...
+from m5.proxy import Parent
+from m5.util.fdthelper import *
+
+
+class LupioRNG(BasicPioDevice):
+    type = 'LupioRNG'
+    cxx_class='gem5::LupioRNG'
+    cxx_header = 'dev/lupio/lupio_rng.hh'
+    pio_size = Param.Addr(0x1000, "PIO Size")
+    byte_order = Param.ByteOrder("big", "Device byte order")
```
joel-porquet, line 38: I don't think we need this parameter. It seems to be unused anyway..

src/dev/lupio/LupioRNG.py
```diff
@@ -0,0 +1,53 @@
...
+    plic = Param.Plic(Parent.any, "PLIC")
+    lupio_rng_int_id = Param.Int(3, "Interrupt ID for the RNG")
```
joel-porquet, line 40: Is it really the right way of determining the interrupt controller and RNG's interrupt ID? One would think that it should be an info at platform level that the device itself has no reason to know.

Given that the LupIO devices are meant to be used in other platforms (than the RISC-V based one, LupV), it feels weird to have this hardcoded connection to the PLIC as well as the hardcoded INT ID here.

I'm not gem5 export though, so I really don't know if it's the expected approach or not.

src/dev/lupio/lupio_rng.cc
```diff
@@ -0,0 +1,149 @@
...
+    DPRINTF(LupioRNG, "LupioRNG initalized with seed: %d\n", user_seed);
+}
+
+uint64_t
+LupioRNG::lupioRNGRead(uint8_t addr)
+{
+    uint32_t r = 0;
+    switch (addr >> 2) {
+        // returns a random number
+        case LUPIO_RNG_RAND: {
```
joel-porquet, line 64: I don't think you need curly brackets for any of the cases here (nor in this read function or the write function below).

Curly brackets are used to define a block, in case a new local variable needs to be declared. AFAIK, you can declare new local variables anywhere with C++, so it seems unnecessary.

src/dev/lupio/lupio_rng.hh
```diff
@@ -0,0 +1,99 @@
...
+
+#include "debug/LupioRNG.hh"
+#include "dev/io_device.hh"
+#include "params/LupioRNG.hh"
+#include "sim/system.hh"
+
+namespace gem5
+{
+
+using namespace RiscvISA;
```
joel-porquet, line 42: Questions:
- Why does a device need to be part of a ISA namespace?
- Can you make it work with multiple ISA namespaces?

src/dev/lupio/lupio_rng.hh
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * LupioRNG:
+ * A Random Number Generator virtual device that returns either a random
+ * value, or a seed that can be configured by the user.
+ */
+class Platform;
+
+class Plic;
```
joel-porquet, line 51: Related to question above: is there a way not to be tied to the PLIC specifically? Can't devices see the interrupt controller they're attached to as a generic component that offers the necessary features?

src/dev/riscv/LupV.py
```diff
@@ -25,13 +25,13 @@
...
 from m5.objects.Clint import Clint
+from m5.objects.LupioRNG import LupioRNG
 from m5.objects.LupioRTC import LupioRTC
 from m5.objects.Platform import Platform
 from m5.objects.Plic import Plic
 from m5.objects.Terminal import Terminal
 from m5.objects.Uart import Uart8250
-from m5.params import Param
-from m5.params import AddrRange
+from m5.params import *
```
joel-porquet, line 34: I was under the impression that it was better to list exactly what needs to be imported, rather than to do blanket imports. Can you confirm with Jason?

src/dev/riscv/LupV.py
```diff
@@ -171,6 +177,12 @@ def attachOffChipIO(self, bus):
             device.pio = bus.mem_side_ports
 
     def generateDeviceTree(self, state):
+        chosen_node = FdtNode("chosen")
+        chosen_node.append(FdtPropertyWords("kernel-start", [0x80000000]))
+        chosen_node.append(FdtPropertyStrings("stdout-path",
+                            "/soc/doesnotexist@0"))
+        yield chosen_node
+
```
joel-porquet, line 185: The chosen node is unrelated to the RNG component. This should be part of another PR.

src/dev/riscv/LupV.py
```diff
@@ -182,11 +194,15 @@ def generateDeviceTree(self, state):
...
+
         for subnode in self.recurseDeviceTree(local_state):
             node.append(subnode)
-
         yield node
 
+
+
+
```
joel-porquet, line 205: Is there a reason for the insertions/removal of empty lines above?

joel-porquet (changes requested):

src/dev/lupio/LupioRNG.py
powerjg: There aren't any good ways to do this right now given the current mess which is the device tree generation. We can either make the device have *known* parameters which are used to generate the device tree, or guess that certain attributes will exist when generating the device tree. IMO, the former is better than the latter. I have Melissa one other suggestion to try. Let us know what you think (assuming it works :))

src/dev/lupio/lupio_rng.hh
mkjost0: I'm not sure of the reasoning behind using this namespace, but as far as I'm aware many of the files within gem5 use this namespace, so we followed that convention.

src/dev/lupio/LupioRNG.py
joel-porquet: I guess this has been fixed since invalid accesses are now properly caught. You can probably resolve the conversation :)

src/dev/lupio/lupio_rng.hh
joel-porquet: Follow-up:

Does the RNG need to have a handle on the PLIC or could it just have a handle on a `BasicPioDevice`? In the latter case, it would make it compatible with any interrupt controller that derives from `BasicPioDevice` (which seems to be the case for all memory-mapped platforms).

Could you try (in that case, instead of declaring `Plic *plic`, you would just declare `BasicPioDevice *pic`), or investigate if it's possible?

src/dev/lupio/LupioRNG.py
```diff
@@ -0,0 +1,58 @@
...
+class LupioRNG(BasicPioDevice):
+
+    type = 'LupioRNG'
+    cxx_class='gem5::LupioRNG'
+    cxx_header = 'dev/lupio/lupio_rng.hh'
+    pio_size = Param.Addr(0x1000, "PIO Size")
+    seed = Param.Int(0, "Initial seed for the random number generator")
+    if buildEnv['TARGET_ISA'] == 'riscv':
+        plic = Param.Plic(Parent.any, "PLIC")
+        lupio_rng_int_id = Param.Int(3, "Interrupt ID for the RNG")
```
joel-porquet, line 42: It seems like the interrupt ID should almost always be given by the platform code when building the hardware system. This value is just a default value because we need to give one. In that case, I would suggest that this default value be 0.

src/dev/lupio/lupio_rng.cc
```diff
@@ -0,0 +1,140 @@
...
+{
+    mt.seed(user_seed);
+    DPRINTF(LupioRNG, "LupioRNG initalized with seed: %d\n", user_seed);
+}
+
+uint64_t
+LupioRNG::lupioRNGRead(uint8_t addr)
+{
+    uint32_t r = 0;
+    std::uniform_int_distribution<uint32_t> distrib(0);
```
joel-porquet, line 61: You usually want to have an empty line between the declaration of your local variables, and the first line of actual code.

src/dev/lupio/lupio_rng.hh
```diff
@@ -0,0 +1,99 @@
...
+
+class LupioRNG : public BasicPioDevice
+{
+  protected:
+    const ByteOrder byteOrder = ByteOrder::big;
```
joel-porquet, line 56: Why big endian?

src/dev/riscv/LupV.py
```diff
@@ -96,6 +103,11 @@ class LupV(Platform):
     # PLIC
     plic = Param.Plic(Plic(pio_addr=0xc000000), "PLIC")
 
+    # LUPIO RNG
+    lupio_rng = LupioRNG(pio_addr=0x20005000)
+    # Interrupt ID for PLIC
+    lupio_rng_int_id = Param.Int(0x3, "PLIC LupioRNG interrupt ID")
+
```
joel-porquet, line 110: I'm bugging on this code a little bit. Shouldn't the interrupt ID be a parameter to the RNG device upon creating it?

At the moment, it seems like the interrupt ID is internally hardcoded in the RNG device and hardcoded here, which means that if they don't match, the code likely breaks :/

joel-porquet (changes requested):
We're getting there, but I still have a few questions about how to make the RNG device as independent from the platform as possible (ie, not dependent from the PLIC specifically, and not tied to a hardcoded interrupt ID).

joel-porquet (approved):
Hopefully I didn't miss anything, but it looks good to me!

PR #229, opened 2021-08-18
WIP: Added new LupIO-TTY device 

lhinman19:


src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,176 @@
...
+     }
+}
+
+LupioTTY::LupioTTY(const Params &params)
+    : BasicPioDevice(params, params.pio_size), device(params.device),
+    byteOrder(params.byte_order),
+    platform(params.platform)
+{
+    // initialize internal registers
+    writ_char = read_char = -1;
```
powerjg, line 54: I wouldn't have multiple assignments on a single line. Also, for class member variables the initialization should be in the initializer list. OR, you can actually initialize in the header if the initial value is a constant.

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,176 @@
...
+uint64_t
+LupioTTY::lupioTTYRead(uint8_t addr)
+{
+    uint32_t r = LUPIO_TTY_INVAL;
+    if (device->dataAvailable()){
+        if (read_char != -1)
+            fprintf(stderr, "Dropping characters\n");
+        else {
+            read_char = device->readData();     //read from the terminal
+            lupio_tty_update_irq();
```
powerjg, line 70: Maybe just have this call once at the end of the function?

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    switch (addr >> 2) {
+        case LUPIO_TTY_WRIT:
+            if (writ_char != -1) {
+                DPRINTF(LupioTTY, "Accessing write register\n");
+                r = writ_char;
```
powerjg, line 78: What is `r`? This variable name is too short (and it's unclear if it's a local variable or class variable).

powerjg (commented):
A few quick comments :)

src/dev/riscv/LupV.py
```diff
@@ -175,6 +183,11 @@ def generateDeviceTree(self, state):
         cpus_node.append(FdtPropertyWords("timebase-frequency", [10000000]))
         yield cpus_node
 
+        chosen_node = FdtNode("chosen")
+        chosen_node.append(FdtPropertyStrings("stdout-path",
```
lhinman19, line 187: I still have been unable to get this to work with "earlycon console=ttyLIO0". As Jason mentioned, if I use kernel parameters "earlycon=lupio_tty,0x20007000" then the earlyconsole does work.

src/dev/lupio/LupioTTY.py
```diff
@@ -0,0 +1,55 @@
...
+    pio_size = Param.Addr(0x1000, "PIO size")
+    platform = Param.Platform(Parent.any,
+                              "Platform this device is part of.")
+
+    def generateDeviceTree(self, state):
+        node = self.generateBasicPioDeviceNode(state, "lupio-tty",
+                                               self.pio_addr,
+                                               self.pio_size)
+        platform = self.platform.unproxy(self)
+        plic = platform.plic
```
joel-porquet, line 49: Back to our convo from the other day, it does seem like this unproxying approach offers a certain level of genericity.

If the platform defined its interrupt controller with a more generic name (eg `intc` or `pic`), then this component (and same for the RNG) could have a line such as `pic = platform.pic`.

The only constraint for using the lupio devices from any platform would then to be that the platform defines a member called `pic`, but the interrupt controller itself could be of any kind (instead of implicitly restricting it to be a PLIC necessarily like the naming here strongly suggests).

In other words, until the new FDT code lands, I suggest to use this unproxying approach but to replace `plic` with `pic`.

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,201 @@
...
+    device->regInterfaceCallback([this]() { dataAvailable(); });
+}
+
+void
+LupioTTY::dataAvailable()
+{
+      // unsure about these checks
+      // if interrupt enable bit is set and the read register is
+      // empty, then can raise an interrupt to receive data
+      if (readIntrEn && readChar == -1) {
```
joel-porquet, line 61: You should actually read the data here: `readChar = device->readData();`. That's what makes everything else work (e.g., returning a correct value for the STAT register)

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,201 @@
...
+          DPRINTF(LupioTTY, "Data available, interrupting\n");
+          platform->postConsoleInt();
```
joel-porquet, line 63: Call `lupioTTYUpdateIRQ()` from here, that's what this function is for.

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,201 @@
...
+{
+    unsigned int irq;
+
+    irq = (writIntrEn && writChar != -1)
+        || (readIntrEn && readChar != -1);
+    if (irq) {
+        DPRINTF(LupioTTY, "LupioTTY InterEvent, interrupting\n");
+        platform->postConsoleInt();
+    } else {
+//        platform->clearConsoleInt();
```
joel-porquet, line 81: The clearing of the INT should be uncommented.

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,201 @@
...
+    if (device->dataAvailable()) {
+        /* Drop characters */
+        if (readChar != -1)
+            fprintf(stderr, "Dropping characters\n");
+        else {
+            // read data from the terminal
+            readChar = device->readData();
+            lupioTTYUpdateIRQ();
+        }
+    }
```
joel-porquet, line 100: Drop that entire block.

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,201 @@
...
+
+    switch (addr >> 2) {
+        case LUPIO_TTY_WRIT:
+            if (writChar != -1) {
+                DPRINTF(LupioTTY, "Accessing write register\n");
+                r = writChar;
+                writChar = -1;
+                lupioTTYUpdateIRQ();
+//                platform->clearConsoleInt();
```
joel-porquet, line 109: Remove this line, the call to the update function should take care of it.

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,201 @@
...
+            }
+            break;
+        case LUPIO_TTY_CTRL:
+            DPRINTF(LupioTTY, "Accessing ctrl register\n");
+                        r = 0;
+                        if (writIntrEn)
+                                r |= LUPIO_TTY_WBIT;
+                        if (readIntrEn)
+                                r |= LUPIO_TTY_RBIT;
+            break;
```
joel-porquet, line 127: Indentation.

joel-porquet (changes requested):
Try with what I suggested, I believe the input should work after that. :crossed_fingers: 

Generally, it seems like the more you stick to the same approach as the QEMU implementation, the higher the chances it will work.

src/dev/riscv/LupV.py
```diff
@@ -175,6 +183,11 @@ def generateDeviceTree(self, state):
...
+                            "/soc/lupio-tty@0x20007000"))
```
joel-porquet, line 188: It should technically be "/soc/lupio-tty@20007000" (for some reason, addresses in node names do not contain the `0x` prefix)

src/dev/riscv/lupv.cc
```diff
@@ -50,12 +52,14 @@ LupV::LupV(const Params &params) :
 void
 LupV::postConsoleInt()
 {
+    plic->post(lupioTTYIntID);
     plic->post(uartIntID);
 }
```
joel-porquet, line 57: You can't raise interrupts for both interrupts. Otherwise, the software will be completely confused. If the uart calls `postConsoleInt()` then only the related IRQ at the PLIC level should be raised, and same for the lupio-tty.

You can either define another set of postConsoleInt()/clearConsoleInt() (eg with a different naming) that is just for the lupio-tty, or you need to accept an argument that says specifically which one should be raised.

I would suggest making a pair of generic functions, such as `postPicInt/clearPicInt(int intID)`. If each device knows its own intID, then it can call this pair of functions with it. It will be useful for lupio-blk as well.

src/dev/lupio/LupioTTY.py
powerjg: That's fine with me. Both approaches are awful in their own way. :)

src/dev/lupio/lupio_tty.cc
powerjg: And curly braces... I wish I could change our style to *require* them. I'm happy to show you examples of how skipping curly braces caused future bugs :)

src/dev/riscv/lupv.cc
powerjg: Something I don't understand about the platform code in general: Is there a reason the lupioTTY has to call into the platform to post the interrupt to the plic? Could it just have a pointer to a generic device and call the post() function itself?

joel-porquet: I guess one drawback of having the device call the `post` function of the interrupt controller directly is that the device needs a pointer to the interrupt controller instance with a type that defines the `post` function. In that case, we're back to the problem of the device being dependent on a certain model of interrupt controller.

A possible solution that may solve this is to have a proper hierarchy of classes. If interrupt controllers derived from a generic interrupt controller class that defines a pair of `post/clear` virtual functions, then devices would only need to have a pointer onto an instance of this generic class and call these functions without any issues.

Or does the BasicPioDevice already define a `post()` function?

powerjg: Ah, that makes sense. No, it doesn't look like there's a virtual `post()` function, which is unfortunate.

I guess you need to add another `LupV::post???Interrupt` function for the lupio TTY. Maybe use `postUartInt()` instead of "console"?

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,201 @@
...
+void
+LupioTTY::lupioTTYUpdateIRQ()
+{
+    unsigned int irq;
+
+    irq = (writIntrEn && writChar != -1)
+        || (readIntrEn && readChar != -1);
+    if (irq) {
+        DPRINTF(LupioTTY, "LupioTTY InterEvent, interrupting\n");
+        platform->postConsoleInt();
```
powerjg, line 79: I don't understand how this can possibly be "generic" between ISAs right now. The platform is ISA (or really *board*) specific. The generic platform (`src/dev/platform.hh`) only has a `postConsoleInterrupt` and a `postPciInt`. I don't know how you're supposed to use these functions.

src/dev/riscv/lupv.cc
lhinman19: Sorry, I left the uart in there because I was switching out the uart and LupIO-TTY when testing and I had intended to remove the uart entirely from the LupV platform. Should I then support both the uart8250 and LupIO-TTY in the LupV platform?

powerjg: Yeah, that's fine for the LupV platform, but now I'm curious as to how this is supposed to work more generally! What if there's two consoles? (I don't think that's impossible in real life...)

src/dev/lupio/lupio_tty.cc
lhinman19: I'm a little concerned that when I clear the INT here, in the Linux boot-up phase I'm getting many "clear interrupt" statements, seemingly without having posted any interrupts. However, everything does seem to work properly and when I was looking at the uart8250 it also seems to output several `cleared` statements without the corresponding interrupt statements.

src/dev/lupio/LupioTTY.py
lhinman19: Should I still make this change now that we have the new FDT code?

powerjg: Yeah, the new code is going to take a few days to get in, and then you'll have to rebase. I vote that you shouldn't be blocked by that :)

src/dev/riscv/lupv.hh
```diff
@@ -45,8 +46,8 @@ class LupV : public Platform
   public:
     Clint *clint;
     LupioRTC *lupioRTC;
-    Plic *plic;
-    int uartIntID;
+    int lupioTTYIntID;
+    Plic *pic;
```
lhinman19, line 50: I'm not sure if this is what was requested. The pic is still a type `Plic` because I wasn't sure if I was supposed to create a generic class named `Pic` from which the `plic` can be derived?

src/dev/lupio/LupioTTY.py
```diff
@@ -0,0 +1,54 @@
...
+from m5.util.fdthelper import *
+
+from m5.objects.Device import BasicPioDevice
+from m5.objects.Serial import SerialDevice
+
+class LupioTTY(BasicPioDevice):
+    type = 'LupioTTY'
+    cxx_class = 'gem5::LupioTTY'
+    cxx_header = "dev/lupio/lupio_tty.hh"
+    device = Param.SerialDevice(Parent.any, "The terminal")
```
joel-porquet, line 38: We may want to find a better name than `device` for this parameter (and then, for the class member). What about `serial` or `terminal`?

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,197 @@
...
+    } else {
+        DPRINTF(LupioTTY, "LupioTTY InterEvent, clear interrupt\n");
+        platform->clearConsoleInt();
+    }
+}
+
+void
+LupioTTY::dataAvailable()
+{
+    if (device->dataAvailable()) {
```
joel-porquet, line 75: I wonder if that should really be an `if`. If this function is called, it *should* be because there is data available. The lack of data would then be an error, caught as such:

```C
assert(device->dataAvailable());
```

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,197 @@
...
+
+    irq = (writIntrEn && writChar != -1)
+        || (readIntrEn && readChar != -1);
+
+    if (irq) {
+        DPRINTF(LupioTTY, "LupioTTY InterEvent, interrupting\n");
+        platform->postConsoleInt();
+    } else {
+        DPRINTF(LupioTTY, "LupioTTY InterEvent, clear interrupt\n");
+        platform->clearConsoleInt();
```
joel-porquet, line 68: The reason for the "clear interrupt" to be printed when the interrupt is actually not *unset* (ie go from high to low) is that this function may be called to operate an update that ends up not actually changing the state of the IRQ. It's more the printf that doesn't report the correct event. You can either remove the printf, or change what it says in order to remove any confusion.

src/dev/riscv/LupV.py
```diff
@@ -94,35 +94,35 @@ class LupV(Platform):
...
 
-    # Uart
-    uart = Uart8250(pio_addr=0x10000000)
+    # LUPIO TTY
+    lupio_tty = LupioTTY(pio_addr=0x20007000)
+
     # Int source ID to redirect console interrupts to
-    # Set to 0 if using a pci interrupt for Uart instead
-    uart_int_id = Param.Int(0xa, "PLIC Uart interrupt ID")
+    lupio_tty_int_id = Param.Int(0xb, "PLIC LupioTTY interrupt ID")
```
joel-porquet, line 106: Why not keep the same interrupt ID as the uart? (it's a question)

src/dev/riscv/LupV.py
```diff
@@ -94,35 +94,35 @@ class LupV(Platform):
...
     terminal = Terminal()
 
-    # Dummy param for generating devicetree
+     # Dummy param for generating devicetree
     cpu_count = Param.Int(0, "dummy")
```
joel-porquet, line 110: Extra space?

src/dev/riscv/lupv.cc
```diff
@@ -42,33 +43,33 @@ LupV::LupV(const Params &params) :
...
-    plic->clear(uartIntID);
+    pic->clear(lupioTTYIntID);
 }
 
 void
 LupV::postPciInt(int line)
 {
-    plic->post(line);
+    pic->post(line);
 }
```
joel-porquet, line 67: In a separate PR, we could probably remove everything related to PCI.

src/dev/riscv/lupv.hh
joel-porquet: It'll get replaced by the lupio-pic eventually, so it's not a big deal.

joel-porquet (changes requested):
Great progress! Does it all work in the end?

src/dev/lupio/lupio_tty.cc
powerjg: `chatty_assert` is preferred: https://gem5.googlesource.com/public/gem5/+/refs/heads/develop/src/base/logging.hh#304

src/dev/riscv/LupV.py
lhinman19: No good reason, except when I was examining the device tree in QEMU I think the interrupt id for lupio-tty was 0xb but I just changed it back to 0xa

lhinman19:
Yes, I was having problems running this code with `telnet` but once I switched over to the `m5term` everything has been working just fine :)

src/dev/lupio/lupio_tty.cc
powerjg: Actually, now it's just `gem5_assert` if you rebase on current develop: https://gem5-review.googlesource.com/c/public/gem5/+/48606

src/dev/lupio/lupio_tty.cc
lhinman19: I think r stands for return value? I can definitely change this naming but we may also need to change it in the RNG and RTC since I think they also use "r".

src/dev/lupio/lupio_tty.cc
joel-porquet: The generic platform does seem to have a virtual function called `postConsoleInt()` (https://github.com/darchr/gem5/blob/develop/src/dev/platform.hh#L62). So I guess calling `platform->postConsoleInt()` should work regardless of the platform, no?

src/dev/lupio/lupio_tty.cc
joel-porquet: Do you require braces for single statements as well?

src/dev/lupio/lupio_tty.cc
joel-porquet: FWIW, `r` is the variable name typically used in QEMU for returning a value in devices' read function. It probably stands for "return value" or "register". That's the name I also used in my QEMU models. If it doesn't fit gem5's coding style, we could use something like `ret`.

joel-porquet (commented):
Looks good to me!

src/dev/lupio/lupio_tty.cc
powerjg: I personally strongly prefer it. Skipping the braces has caused hard to find bugs in the past.

src/dev/lupio/lupio_tty.cc
powerjg: My confusion is which console should this post the interrupt to? What if there are two consoles (i.e., in this case with both the uart and the lupioTTY

joel-porquet: I guess the current model doesn't allow this scenario. If you wanted to allow it, you'd probably need for `postConsoleInt()` to accept an ID of some sort to indicate which console is raising its IRQ.

That said, the lupio-TTY is replacing the uart now, so we actually don't need the support for two consoles.

lhinman19:
I just rebased so that it now also includes the lupIO-RNG.

src/dev/lupio/lupio_tty.cc
```diff
@@ -0,0 +1,197 @@
...
+            // write data to terminal
+            terminal->writeData(c);
+            writChar = c;
+            lupioTTYUpdateIRQ();
+            break;
+        case LUPIO_TTY_CTRL:
+            // set interrupt enable bits
+            DPRINTF(LupioTTY, "Accessing LUPIO_TTY_CTRL\n");
+            writIntrEn = !!(val & LUPIO_TTY_WBIT);
+            readIntrEn = !!(val & LUPIO_TTY_RBIT);
```
powerjg, line 156: You shouldn't need the `!!` (unless I'm mistaken). The cast from `uint8_t` to `bool` should implicitly work.

```suggestion
            writIntrEn = val & LUPIO_TTY_WBIT;
            readIntrEn = val & LUPIO_TTY_RBIT;
```

powerjg (commented):
One more tiny thing, then it's good to go from my end.

src/dev/lupio/lupio_tty.cc
joel-porquet: It's a pretty common idiom in the Linux kernel, which I think comes from the fact that the kernel doesn't use `bool` very often (but `int` instead). Here you're right, the conversion to bool should work.

PR #231, opened 2021-09-08
Added new LupV Board for LupIO Devices

mkjost0:


components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,342 @@
...
+        self.workload = RiscvLinux()
+        # Contains a CLINT, PLIC, UART, and some functions for the dtb, etc.
+        self.platform = LupV()
+        # Note: This only works with single threaded cores.
+        self.platform.pic.n_contexts = self.processor.get_num_cores() * 2
+        self.platform.attachPlic()
+        self.platform.clint.num_threads = self.processor.get_num_cores()
+        # Add the RTC
+        # TODO: Why 100MHz? Does something else need to change when this does?
+        self.platform.rtc = RiscvRTC(frequency=Frequency("100MHz"))
```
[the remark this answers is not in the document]
mkjost0: I think this RTC helps handle interrupts in a way that the LupIO-RTC doesn't, so removing it from the board causes the block device to hang.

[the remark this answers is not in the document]
powerjg: Aha!

Interesting...

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,342 @@
...
+        # CLINT node
+        clint = self.platform.clint
+        clint_node = clint.generateBasicPioDeviceNode(
+            soc_state, "clint", clint.pio_addr, clint.pio_size
+        )
+        int_extended = list()
+        for i, core in enumerate(self.get_processor().get_cores()):
+            phandle = soc_state.phandle(f"cpu@{i}.int_state")
+            int_extended.append(phandle)
+            int_extended.append(0x3)
```
joel-porquet, line 265: Can we use named constant in Python? (would be better than hardcoded values)

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,342 @@
...
+        # Default DTB address if bbl is built with --with-dts option
+        self.workload.dtb_addr = 0x87E00000
+        # We need to wait to generate the device tree until after the disk is
+        # set up. Now that the disk and workload are set, we can generate the
+        # device tree file.
+        self.generate_device_tree(m5.options.outdir)
+        self.workload.dtb_filename = os.path.join(
+            m5.options.outdir, "device.dtb"
+        )
+    def generate_device_tree(self, outdir: str) -> None:
```
joel-porquet, line 193: Can you add some empty spaces inside your functions to separate the different blocks of code? It's pretty dense and hard to read otherwise.

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,342 @@
...
+    This board assumes that you will be booting Linux.
+    **Limitations**
+    * Only works with classic caches
+    """
+    def __init__(
+        self,
+        clk_freq: str,
+        processor: AbstractProcessor,
+        memory: AbstractMemorySystem,
+        cache_hierarchy: AbstractCacheHierarchy,
```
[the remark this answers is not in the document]
powerjg: I'm going to mark this as resolved. I talked to Bobby about it today and he convinced me that this is the right way to do it :)

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,394 @@
...
+    ) -> None:
+        super().__init__(clk_freq, processor, memory, cache_hierarchy)
+        if get_runtime_isa() != ISA.RISCV:
+            raise EnvironmentError(
+                "RiscvBoard will only work with the RISC-V ISA. Please"
+                " recompile gem5 with ISA=RISCV."
+            )
+        if cache_hierarchy.is_ruby():
+            raise EnvironmentError("RiscvBoard is not compatible with Ruby")
```
powerjg, line 89: These may be able to be replaced with `requires()` calls depending on if that's been merged or not.

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,394 @@
...
+        self.lupio_rtc = LupioRTC(pio_addr=0x20004000)
+
+        # LUPIO TTY
+        self.lupio_tty = LupioTTY(pio_addr=0x20007000, pic = self.pic)
+        self.terminal = Terminal()
+
+        # LUPIO BLK
+        self.lupio_blk = LupioBLK(pio_addr=0x20000000, pic = self.pic)
+
+        # Note: This only works with single threaded cores.
```
powerjg, line 111: Sadly, you can't use `requires()` here. @BobbyRBruce, it would be nice if you could say `requires(self.processor.get_num_cores() == 1)` or something like that, right?

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,394 @@
...
+        self.pic.n_contexts = self.processor.get_num_cores() * 2
+        # Interrupt IDs for LupIO devices
+        pic_srcs = [
+            self.lupio_tty.lupio_tty_int_id,
+            self.lupio_blk.int_id,
+            self.lupio_rng.lupio_rng_int_id
```
powerjg, line 117:
```suggestion
            self.lupio_tty.int_id,
            self.lupio_blk.int_id,
            self.lupio_rng.int_id
```

You can remove the `lupio_tty` parts of the names, right?

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,394 @@
...
+        tested kernels and OSes.
+        The command is an optional string to execute once the OS is fully
+        booted, assuming the disk image is setup to run `m5 readfile` after
+        booting.
+        After the workload is set up, this functino will generate the device
+        tree file and output it to the output directory.
+        **Limitations**
+        * Only supports a Linux kernel
+        * Disk must be configured correctly to use the command option
+        * This board doesn't support the command option
```
powerjg, line 200: Make sure this is up to date

components_library/boards/lupv_board.py
powerjg: Agreed. Where do these numbers come from? If it's part of the board, then we can reference a board variable, too.

src/dev/lupio/LupioBLK.py
```diff
@@ -41,23 +41,5 @@ class LupioBLK(DmaDevice):
...
-        platform = self.platform.unproxy(self)
-        pic = platform.pic
-        node.append(
-                FdtPropertyWords("interrupts", [platform.lupio_blk_int_id]))
-        node.append(
-                FdtPropertyWords("interrupt-parent", state.phandle(pic)))
-        node.appendCompatible(["lupio,blk"])
-        yield node
-
+    pic = Param.Plic("PLIC")
```
powerjg, line 44: I think this is the right solution for now. However, we should talk about how to generalize the interrupt controller interface so the block device isn't RISC-V specific.

This would require refactoring a bunch of code in gem5, so it isn't important right now.

src/dev/lupio/LupioTTY.py
```diff
@@ -37,18 +37,5 @@ class LupioTTY(BasicPioDevice):
...
-        platform = self.platform.unproxy(self)
-        pic = platform.pic
-        node.append(
-            FdtPropertyWords("interrupts", [platform.lupio_tty_int_id]))
-        node.append(
-                FdtPropertyWords("interrupt-parent", state.phandle(pic)))
-        node.appendCompatible(["lupio,tty"])
-        yield node
+    pic = Param.Plic("PLIC")
+    lupio_tty_int_id = Param.Int(0xa, "interrupt")
```
powerjg, line 41:
```suggestion
    int_id = Param.Int(0xa, "The interrupt number for this device, default works for RISC-V")
```

I'd add some more details to the description. I'm not sure if my details are correct.

src/dev/lupio/lupio_tty.cc
```diff
@@ -46,7 +46,8 @@ namespace gem5
 LupioTTY::LupioTTY(const Params &params)
     : BasicPioDevice(params, params.pio_size),
     writChar(-1), readChar(-1), writIntrEn(false), readIntrEn(false),
-    terminal(params.terminal), platform(params.platform)
+    terminal(params.terminal), pic(params.pic),
+    lupioTTYIntID(params.lupio_tty_int_id)
```
powerjg, line 50:
```suggestion
    intID(params.int_id)
```

This is in the LupioTTY class. There's no need to repeat that in the variable name.

components_library/boards/lupv_board.py
mkjost0: How would I use the `requires()` call?

powerjg: Here's an example: https://gem5.googlesource.com/public/gem5/+/2b073505b532fdd4b15298d738be8f80cb9c96c7/src/python/gem5/components/boards/riscv_board.py#90

You can see more in this review request: https://gem5-review.googlesource.com/c/public/gem5/+/49867/13

joel-porquet:
From what I understand, the (new) third commit modifies some of what the first commit is doing. Would it makes sense to fold some of this commit (or even the entire thing?) into the first commit? Otherwise, it's a bit difficult to review this PR knowing that the last commit fixes some aspects of the first commit.. Thoughts?

powerjg:
We could squash them all.. I've been just looking at the final changes and not per-commit, personally.

I'm not sure if it matters too much right now. We will probably have to merge these into other changesets in the lupio/main branch before pushing to upstream anyway. I don't think we want upstream to first see the version with the lupv platform and then removing the platform.

I could be wrong though.

components_library/boards/lupv_board.py
```diff
@@ -0,0 +1,389 @@
...
+        return True
+    def get_mem_side_coherent_io_port(self) -> Port:
+        return self.iobus.mem_side_ports
+    @overrides(AbstractBoard)
+    def setup_memory_ranges(self):
+        memory = self.get_memory()
+        mem_size = memory.get_size()
+        self.mem_ranges = [AddrRange(start=0x80000000, size=mem_size)]
+        memory.set_memory_range(self.mem_ranges)
+    def set_workload(
```
BobbyRBruce, line 186: I think Python Black needs run here. There should be a space between these functions. Not a big deal but it looks a little ugly.

BobbyRBruce (commented):
Looks good to me. I see no major problems. Really good work.

I think we should create a new directory under boards called `experimental`, where this board will live.

PR #232, opened 2021-09-13
WIP: Error in initializing LupIO-TMR

lhinman19:


components_library/boards/lupv_board.py
```diff
@@ -312,6 +335,44 @@ def generate_device_tree(self, outdir: str) -> None:
         clint_node.appendCompatible(["riscv,clint0"])
         soc_node.append(clint_node)
 
+        # Clock
+        clk_node = FdtNode("lupv-clk")
+        clk_phandle = state.phandle(f"cpu@{0}.int_state")
```
joel-porquet, line 340: The clock needs to generate its own phandle, not reuse the CPU's phandle.

Try something like `clk_phandle = state.phandle(clk_node)` to generate a new phandle that's related to this clock node. Then in `lupio-tmr`, you can refer back to it.

lhinman19: When I try `clk_phandle = state.phandle(clk_node)`, the `lupv-clk`'s phandle becomes `<0x00000003>` and I am still getting that error: `could not get #clock-cells for /lupv-clk`

PR #233, opened 2021-09-14
WIP: Added LupIO-TMR and LupIO-PIC

lhinman19:


components_library/boards/lupv_board.py
```diff
@@ -320,21 +362,39 @@ def generate_device_tree(self, outdir: str) -> None:
         int_state = FdtState(addr_cells=0, interrupt_cells=1)
         plic_node.append(int_state.addrCellsProperty())
         plic_node.append(int_state.interruptCellsProperty())
-        phandle = int_state.phandle(plic)
+        phandle = state.phandle(plic)
         plic_node.append(FdtPropertyWords("phandle", [phandle]))
-        plic_node.append(FdtPropertyWords("riscv,ndev", [self.pic.n_src - 1]))
+        plic_node.append(FdtPropertyWords("riscv,ndev", 0))
```
powerjg, line 367: Was this wrong before? Can you explain this change a bit? (I mostly just want to understand the device tree better)

src/dev/lupio/LupioPIC.py
```diff
@@ -0,0 +1,40 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects.Device import BasicPioDevice
+from m5.params import Param
+#from m5.defines import buildEnv
```
powerjg, line 29: delete unneeded lines

src/dev/lupio/lupio_tmr.hh
```diff
@@ -0,0 +1,115 @@
...
+ * LupioTMR:
+ * A timer virtual device 
+ */
+class LupioPIC;
+
+class LupioTMR : public BasicPioDevice
+{
+  protected:
+    const ByteOrder byteOrder = ByteOrder::little;
+    System *system;
```
powerjg, line 51: Is  this used?

src/dev/lupio/lupio_tmr.hh
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * LupioTMR:
+ * A timer virtual device 
+ */
+class LupioPIC;
+
+class LupioTMR : public BasicPioDevice
+{
+  protected:
```
powerjg, line 49: Why not private?

src/dev/lupio/lupio_tmr.hh
```diff
@@ -0,0 +1,115 @@
...
+    const ByteOrder byteOrder = ByteOrder::little;
+    System *system;
+    EventFunctionWrapper tmrEvent;
+
+    Tick start = 0; 
+    Tick next = 0; 
+    Cycles startCycle = curCycle();
```
powerjg, line 56: 1. I'd put this in the initializer list with the constructor. I think it's OK to use const initialization in the header, but non const initialization scares me a bit.
2. When restoring from a checkpoint, this isn't going to work. If you really need this to be the current cycle at the beginning of simulation, you should grab it in `startup`. See http://doxygen.gem5.org/release/current/classgem5_1_1SimObject.html#details for details.

src/dev/lupio/lupio_tmr.hh
```diff
@@ -0,0 +1,115 @@
...
+
+    // Timer registers
+    uint64_t reload = 0;
+
+    // Control
+    bool ie = false;
+    bool pd = false;
+
+    // Status
+    bool expired =false;
```
powerjg, line 82:
```suggestion
    bool expired = false;
```

src/dev/lupio/lupio_tmr.hh
```diff
@@ -0,0 +1,115 @@
...
+
+  protected:
```
powerjg, line 84: Why protected and not private?

src/dev/lupio/lupio_tmr.hh
```diff
@@ -0,0 +1,115 @@
...
+     * Function to launch or stop the timer depending on the load value
+     */
+    void lupioTMRWrite(const uint8_t addr, uint64_t val64, int size);
+    
+    uint64_t lupioTMRCurrentTime();
+    void lupioTMRSet();
+    void lupioTMRCallback();
+
+    void updateIRQ(int level);
+
```
powerjg, line 100: Do these need documentation, or are they obvious?

components_library/boards/lupv_board.py
mkjost0: To my understanding, we set it as 0 here because now the LupIO-PIC handles all the devices, so the Plic can be set to 0.  However, we made this change because it was one of the steps in the README we followed.

powerjg: I see! Thanks!

src/dev/lupio/lupio_tmr.hh
mkjost0: We used system so that the timer could directly update interrupt requests in a similar fashion to the CLINT, since based on the QEMU implementation, the LupIO-TMR doesn't have its own interrupt ID to connect to the LupIO-PIC.  However, we were wondering whether it made sense to keep it like this, or to adapt the LupIO-TMR to update its interrupt requests through the LupIO-PIC like all the other devices do.

powerjg: Ah, yes, I see now. I'll comment there.

src/dev/lupio/lupio_tmr.cc
```diff
@@ -0,0 +1,199 @@
...
+    system(params.system),
+    tmrEvent([this]{ lupioTMRCallback(); }, name()) 
+{
+    start = curTick(); 
+    DPRINTF(LupioTMR, "LupioTMR initalized\n");
+}
+
+void LupioTMR::updateIRQ(int level)
+{
+    auto tc = system->threads[0];
```
powerjg, line 57: I don't think you can always assume that system->threads[0] is what you want. Let's talk about a better solution in our meeting.

src/dev/lupio/lupio_tmr.cc
```diff
@@ -0,0 +1,199 @@
...
+
+    switch (addr >> 2) {
+        case LUPIO_TMR_LOAD:
+                reload = val;
+                DPRINTF(LupioTMR, "Write LUPIO_TMR_LOAD: %d\n", reload);
+            break;
+
+        case LUPIO_TMR_CTRL:
+			ie = val & LUPIO_TMR_IE;
+			pd = val & LUPIO_TMR_PD;
```
joel-porquet, line 145: The indentation is not consistent for the lines above. Probably a mix of spaces and tabs.

src/dev/lupio/lupio_pic.cc
```diff
@@ -0,0 +1,171 @@
...
+    lupioPicUpdateIRQ();
+}
+
+void
+LupioPIC::lupioPicUpdateIRQ()
+{
+    for (int i = 0; i < nContext; i++) {
+        int thread_id = i >> 1;
+        int int_id = (i & 1) ?
+            ExceptionCode::INT_EXT_SUPER : ExceptionCode::INT_EXT_MACHINE;
```
joel-porquet, line 79: Don't copy from the PLIC, their approach is different and very RISC-V oriented (ie this notion of contexts).

For now, I would strongly suggest making a mono-cpu design, that is N sources (devices connected to PIC) and only 1 output (the CPU).

The type of interrupt (eg int_ext_super) should be a parameter to the device.

src/dev/lupio/lupio_pic.cc
```diff
@@ -0,0 +1,171 @@
...
+            // Value will be 32 if there is no unmasked pending IRQ
+            r = ctz32(pending & mask);
+            DPRINTF(LupioPIC, "Read PIC_PRIO: %d\n", r);
+            break;
+        case LUPIO_PIC_MASK:
+            r = mask;
+            DPRINTF(LupioPIC, "Read PIC_MASK: %d\n", r);
+            break;
+        case LUPIO_PIC_PEND:
+			r = pending;
```
joel-porquet, line 108: Inconsistent indentation.

components_library/boards/lupv_board.py
```diff
@@ -95,28 +97,39 @@ def __init__(
...
 
         # LUPIO BLK
-        self.lupio_blk = LupioBLK(pio_addr=0x20000000, pic = self.pic)
+        self.lupio_blk = LupioBLK(pio_addr=0x20000000, pic = self.lupio_pic)
 
         # Note: This only works with single threaded cores.
         self.pic.n_contexts = self.processor.get_num_cores() * 2
+        
+        self.lupio_pic.n_contexts = self.processor.get_num_cores() * 2
+
```
joel-porquet, line 123: Doesn't apply for us.

components_library/boards/lupv_board.py
```diff
@@ -95,28 +97,39 @@ def __init__(
...
-        self.lupio_tty = LupioTTY(pio_addr=0x20007000, pic = self.pic)
+        self.lupio_tty = LupioTTY(pio_addr=0x20007000, pic = self.lupio_pic)
         self.terminal = Terminal()
 
         # LUPIO BLK
-        self.lupio_blk = LupioBLK(pio_addr=0x20000000, pic = self.pic)
+        self.lupio_blk = LupioBLK(pio_addr=0x20000000, pic = self.lupio_pic)
 
         # Note: This only works with single threaded cores.
         self.pic.n_contexts = self.processor.get_num_cores() * 2
```
joel-porquet, line 120: Actually, now that our LupioPIC is in charge of the supervisor-level external interrupts, the PLIC should be left in charge of the machine-level external interrupts only. So technically, the number of contexts should only be `self.processor.get_num_cores()`.

components_library/boards/lupv_board.py
```diff
@@ -95,28 +97,39 @@ def __init__(
...
+        
+        self.lupio_pic.n_contexts = self.processor.get_num_cores() * 2
+        
         # Interrupt IDs for LupIO devices
         pic_srcs = [
             self.lupio_tty.int_id,
             self.lupio_blk.int_id,
             self.lupio_rng.int_id
         ]
         self.pic.n_src = max(pic_srcs) + 1
```
joel-porquet, line 130: Should be 0.

components_library/boards/lupv_board.py
joel-porquet: Also, BBL can recognize when `riscv,ndev==0` and it returns from PLIC-related functions faster.

joel-porquet (changes requested):

src/dev/lupio/lupio_pic.cc
joel-porquet: Not resolved.

src/dev/riscv/LupV.py
```diff
@@ -0,0 +1,43 @@
...
+from m5.objects.Terminal import Terminal
+from m5.params import Param
+from m5.params import AddrRange
+
+class LupV(Platform):
+    type = 'LupV'
+    cxx_header = "dev/riscv/lupv.hh"
+    cxx_class = 'gem5::LupV'
+    pic = Param.LupioPIC("PIC")
+    lupio_tty_int_id = Param.Int(0xa, "PLIC LupioTTY interrupt ID")
```
joel-porquet, line 43: Remove default value, it *has* to be received.

components_library/boards/lupv_board.py
```diff
@@ -111,11 +115,13 @@ def __init__(
         self.lupio_tmr = LupioTMR(pio_addr=0x20006000)
 
         # LUPIO TTY
-        self.lupio_tty = LupioTTY(pio_addr=0x20007000, pic = self.lupio_pic)
+        self.lupio_tty = LupioTTY(pio_addr=0x20007000, platform = self.lupv,
+                                    int_id = 0xa)
```
joel-porquet, line 119: It'd be great to create a dictionary for the irq number or something, instead of using these hardcoded values that can only create bugs down the line.

```py
int_ids = { 'TTY': 0, 'BLK': 1, 'RNG': 2}
...
self.lupio_tty = LupioTTY(pio_addr=0x20007000, platform = self.lupv, int_id = int_ids['TTY'] )
```

components_library/boards/lupv_board.py
```diff
@@ -101,6 +102,9 @@ def __init__(
         # The interrupt type is INT_EXT_SUPER
         self.lupio_pic = LupioPIC(pio_addr=0x20002000, int_type = 0x9)
```
joel-porquet, line 103: Could this `0x9` be replace by the proper exception number symbol (like you had in the PIC)?

lhinman19: The interrupt codes are defined in `arch/riscv/faults.hh` (https://github.com/darchr/gem5/blob/develop/src/arch/riscv/faults.hh). I'm not sure how to port that over to Python.

joel-porquet: In that case, you could at least define a python dict to associate the value to a relevant name:

```python
excep_code = {'INT_EXT_SUPER': 9, 'INT_EXT_MACHINE': 10, ...}
```

This way you can use the name, which makes the code more self-explanatory.

src/dev/riscv/LupV.py
```diff
@@ -0,0 +1,35 @@
...
+
+from m5.objects.Platform import Platform
+from m5.params import Param
+
+class LupV(Platform):
+    type = 'LupV'
+    cxx_header = "dev/riscv/lupv.hh"
+    cxx_class = 'gem5::LupV'
+    pic = Param.LupioPIC("PIC")
+    lupio_tty_int_id = Param.Int("PLIC LupioTTY interrupt ID")
```
powerjg, line 35: Is this for the uart or the lupv tty?

powerjg (approved):
One small question that I just posted. Otherwise, let's merge this so we can more easily rebase and start thinking about how to post upstream.

src/dev/riscv/LupV.py
mkjost0: It's for the LupV TTY.  It's only there so that if someone calls the `postConsoleInt` function from LupV, it interrupts properly, but we don't actually use it within our implementation of the LupIO-TTY.

PR #236, opened 2021-10-14
WIP: Added SMP Support

lhinman19:
Sorry this took so long! I finally got SMP support to work!

### DIFF
--- a/src/dev/lupio/lupio_ipi.cc
+++ b/src/dev/lupio/lupio_ipi.cc
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2021 The Regents of the University of California
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "dev/lupio/lupio_ipi.hh"
+
+#include "cpu/base.hh"
+#include "debug/LupioIPI.hh"
+#include "mem/packet_access.hh"
+#include "params/LupioIPI.hh"
+
+namespace gem5
+{
+
+LupioIPI::LupioIPI(const Params &params) :
+    BasicPioDevice(params, params.pio_size),
+    system(params.system),
+    intType(params.int_type),
+    nThread(params.num_threads)
+{
+    word.resize(nThread, 0);
+
+    DPRINTF(LupioIPI, "LupioIPI initalized--number of CPUs: %d\n", nThread);
+}
+
+uint64_t
+LupioIPI::lupioIPIRead(uint8_t addr, int size)
+{
+    int cpu = addr >> 2;
+    uint32_t r = 0;
+    // Reading automatically lowers corresponding IRQ
+    r = word[cpu];
+
+    auto tc = system->threads[cpu];
+    tc->getCpuPtr()->clearInterrupt(tc->threadId(), intType, 0);
+    // Also reset value after reading
+    word[cpu] = 0;
+
+    return r;
+}
+
+void
+LupioIPI::lupioIPIWrite(uint8_t addr, uint64_t val64, int size)
+{
+    int cpu = addr >> 2;;
+    uint32_t val = val64;
+
+    word[cpu] = val;
+
+    // Raise IRQ
+    auto tc = system->threads[cpu];
+
+    tc->getCpuPtr()->postInterrupt(tc->threadId(), intType, 0);
+}
+
+Tick
+LupioIPI::read(PacketPtr pkt)
+{
+    Addr ipi_addr = pkt->getAddr() - pioAddr;
+
+    DPRINTF(LupioIPI,
+        "Read request - addr: %#x, size: %#x\n", ipi_addr, pkt->getSize());
+
+    uint64_t read_val = lupioIPIRead(ipi_addr, pkt->getSize());
+    DPRINTF(LupioIPI, "Packet Read: %#x\n", read_val);
+    pkt->setUintX(read_val, byteOrder);
+    pkt->makeResponse();
+
+    return pioDelay;
+}
+
+Tick
+LupioIPI::write(PacketPtr pkt)
+{
+    Addr ipi_addr = pkt->getAddr() - pioAddr;
+
+    DPRINTF(LupioIPI, "Write register %#x value %#x\n", ipi_addr,
+            pkt->getUintX(byteOrder));
+
+    lupioIPIWrite(ipi_addr, pkt->getUintX(byteOrder), pkt->getSize());
+    DPRINTF(LupioIPI, "Packet Write Value: %d\n", pkt->getUintX(byteOrder));
+
+    pkt->makeResponse();
+
+    return pioDelay;
+}
+} // namespace gem5
+
--- a/src/dev/lupio/lupio_ipi.hh
+++ b/src/dev/lupio/lupio_ipi.hh
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2021 The Regents of the University of California
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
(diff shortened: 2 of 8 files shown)


joel-porquet (changes requested):
Nice! I started working on this last night actually and was about to write similar code :P It seems like what was missing was to have an individual timer per cpu.

I suggested an improved design for creating instances of individual timers, rather than having a bunch of separate arrays.

src/dev/lupio/lupio_tmr.cc
```diff
@@ -80,52 +88,63 @@ LupioTMR::lupioTMRCurrentTime()
...
-        schedule(tmrEvent, (reload * sim_clock::as_int::ns) + curTick());
-    }
+    // Start the timer
+    startTime[cpu] = curTick();
+
+    Event* tmr_event = new EventFunctionWrapper(
+        [=]{
+            lupioTMRCallback(cpu);
+        }, name()+"done"
+    );
```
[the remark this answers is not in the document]
lhinman19: I was having trouble getting it to work with the parameter set to "true" and it seemed like the timer was getting deleted prematurely? So now I just have the timer events created in the constructor.

src/python/gem5/components/boards/lupv_board.py
```diff
@@ -94,14 +95,20 @@ def __init__(
 
         self._excep_code = { 'INT_SOFT_SUPER': 1, 'INT_TIMER_SUPER': 5,
                              'INT_TIMER_MACHINE': 7, 'INT_EXT_SUPER': 9,
-                             'INT_EXT_MACHINE': 10 }
+                             'INT_EXT_MACHINE': 11 }
```
joel-porquet, line 98: Not a big big deal but the current formatting of this dict is quite confusing (I struggled with it for a couple minutes when I was working on this code last week). The interrupt number related to the same event type should ideally be on the same line, for symmetry:

```python
        self._excep_code = { 'INT_SOFT_SUPER'  : 1,
                             'INT_TIMER_SUPER' : 5, 'INT_TIMER_MACHINE' : 7,
                             'INT_EXT_SUPER'   : 9, 'INT_EXT_MACHINE'   : 11 }
```

Makes the array much easier to read IMO :)

src/python/gem5/components/boards/lupv_board.py
```diff
@@ -94,14 +95,20 @@ def __init__(
...
         # CLINT
         self.clint = Clint(pio_addr=0x2000000)
 
         # PLIC
         self.pic = Plic(pio_addr=0xc000000)
 
+        # LUPIO IPI
+        self.lupio_ipi = LupioIPI(pio_addr=0x20001000,
+                                  int_type =
+                                  self._excep_code['INT_SOFT_SUPER'],
```
joel-porquet, line 109: Even if it slightly goes over 80 chars, it might be better to merge this line with the previous one. Otherwise, it's kind of hard to read.

joel-porquet (commented):
I'm nitpicking a little, but overall it's a great commit.

src/python/gem5/components/boards/lupv_board.py
lhinman19: There is a style checker that doesn't allow us to go over 80 chars so I wasn't sure what to do about this.

lhinman19:
Is this ok to be merged?

src/dev/lupio/lupio_ipi.cc
```diff
@@ -0,0 +1,125 @@
...
+namespace gem5
+{
+
+LupioIPI::LupioIPI(const Params &params) :
+    BasicPioDevice(params, params.pio_size),
+    system(params.system),
+    intType(params.int_type),
+    nCPUs(params.num_cores)
+{
+    int cpu;
```
powerjg, line 47: This variable should be in the for loop statement below

src/dev/lupio/lupio_ipi.hh
```diff
@@ -0,0 +1,90 @@
...
+
+#ifndef __DEV_LUPIO_LUPIO_IPI_HH__
+#define __DEV_LUPIO_LUPIO_IPI_HH__
+
+#include "arch/riscv/interrupts.hh"
+#include "dev/io_device.hh"
+#include "dev/platform.hh"
+#include "params/LupioIPI.hh"
+
+#define LUPIO_IPI_NCPU      32
```
powerjg, line 37: I assume this is the maximum number of CPUs, right?

Is there a fundamental reason why this is the max? Any reason not to use a std::vector for `word` below?

src/dev/lupio/lupio_ipi.hh
```diff
@@ -0,0 +1,90 @@
...
+        // Max offset
+        LUPIO_IPI_MAX,
+    };
+
+    uint32_t nCPUs;
+
+    uint32_t word[LUPIO_IPI_NCPU];
+
+    /**
+     * Function to return
```
powerjg, line 68: These comments are incomplete.

src/dev/lupio/lupio_pic.cc
```diff
@@ -35,6 +35,7 @@
 #include "sim/system.hh"
 
 #define LUPIO_PIC_NSRC      32
+#define LUPIO_PIC_NCPU      32
```
powerjg, line 38: Is it required that this and LUPIO_IPI_NCPU are the same? If so, then this is a potential future bug if they become out of sync. If there is some fundamental reason why there is a specific maximum number of CPUs, then that should be specified as a single variable. If it's possible to make this a parameter it would be even better.

src/dev/lupio/lupio_pic.cc
```diff
@@ -48,6 +49,11 @@ LupioPIC::LupioPIC(const Params &params) :
     nThread(params.num_threads),
     intType(params.int_type)
 {
+    memset(enable, 0, LUPIO_PIC_NCPU * sizeof(uint32_t));
+    memset(mask, 0, LUPIO_PIC_NCPU * sizeof(uint32_t));
```
powerjg, line 53: I'm pretty sure you can do the following, which is more idiomatic C++:

```suggestion
    enable = {0};
    mask = {0};
```

This is also better in the case that you ever change the type of `enable` or `mask` to `uint64_t`

src/dev/lupio/lupio_pic.cc
```diff
@@ -74,37 +80,45 @@ LupioPIC::clear(int src_id)
 void
 LupioPIC::lupioPicUpdateIRQ()
 {
-    if (nThread > 1 ) {
-        panic("This device currently does not have SMP support\n");
-    }
+    int cpu;
```
powerjg, line 83: Inside the for statement

src/dev/lupio/lupio_pic.cc
```diff
@@ -74,37 +80,45 @@ LupioPIC::clear(int src_id)
...
     }
 }
 
 uint64_t
 LupioPIC::lupioPicRead(uint8_t addr)
 {
     uint32_t r = 0;
-    
-    switch (addr >> 2) {
+    int cpu, reg;
```
powerjg, line 100: Put these declarations on the same line as the initialization. We can use C++-17+, no need to restrict ourselves to C-98

src/dev/lupio/lupio_tmr.cc
```diff
@@ -80,52 +84,59 @@ LupioTMR::lupioTMRCurrentTime()
...
 {
-    startTime = curTick();
-    if (!tmrEvent.scheduled()) {
-        schedule(tmrEvent, (reload * sim_clock::as_int::ns) + curTick());
+    // Start the timer
+    _timers[cpu].startTime = curTick();
+
+    if (!_timers[cpu].tmrEvent->scheduled()) {
+        schedule(*(_timers[cpu].tmrEvent),
+                (_timers[cpu].reload * sim_clock::as_int::ns) + curTick());
```
powerjg, line 94: Can you add a comment on this math?

src/dev/lupio/lupio_tmr.cc
```diff
@@ -80,52 +84,59 @@ LupioTMR::lupioTMRCurrentTime()
...
 }
 
 uint64_t
 LupioTMR::lupioTMRRead(uint8_t addr, int size)
 {
-    uint64_t r = 0;
-    
-    switch (addr >> 2) {
+    uint32_t r = 0;
+    int cpu, reg;
```
powerjg, line 117: Declaration and initialization can be on the same line :)

src/dev/lupio/lupio_tmr.cc
```diff
@@ -140,27 +151,33 @@ void
 LupioTMR::lupioTMRWrite(uint8_t addr, uint64_t val64, int size)
 {
     uint32_t val = val64;
+    int cpu, reg;
```
powerjg, line 154: Declaration and initialization on the same line

src/dev/lupio/lupio_tmr.hh
```diff
@@ -67,15 +65,17 @@ class LupioTMR : public BasicPioDevice
...
+        uint64_t reload = 0;
+        bool ie = false;            // Control
+        bool pd = false;
+        bool expired = false;       // Status
+        Tick startTime = 0;
+    };
 
-    // Status
-    bool expired = false;
+    std::vector<LupioTimer> _timers;
```
powerjg, line 78: Why `_timers` and not `timers`? You don't use leading underscores for private members anywhere else.

src/dev/lupio/lupio_tmr.cc
```diff
@@ -49,26 +49,30 @@ LupioTMR::LupioTMR(const Params &params) :
     BasicPioDevice(params, params.pio_size),
     system(params.system),
     nThread(params.num_threads),
-    tmrEvent([this]{ lupioTMRCallback(); }, name()),
     intType(params.int_type)
 {
+    _timers.resize(nThread);
+
+    for (int cpu = 0; cpu < nThread; cpu++) {
+        _timers[cpu].tmrEvent = new EventFunctionWrapper(
```
powerjg, line 57: Where do you call delete? You should probably add a destructor for either the `LupioTimer` struct or for the `LupioTMR` which deletes this dynamically allocated memory.

powerjg (commented):
Sorry that I haven't looked at this closely sooner! Below are a bunch of comments. It would be OK with me if you wanted to merge this as-is and make these changes before pushing to mainline gem5.

src/dev/lupio/lupio_pic.cc
joel-porquet: It's funny because I started writing this comment to explain why the number of cpus was limited to 32 by design, only to realize that it actually wasn't. Unless I've missed something, it looks like LupIO can support any number of CPUs. Each SMP-ready device (timer, pic, ipi) can have its interface duplicated in as many CPUs as there are in the system.

The number of IRQ sources is limited to 32 by design though, because LupIO assumes 32-bit registers across all devices. The 32-bit `pending` register can therefore reflect IRQs from 32 devices max. That explains the `NSRC` value.

src/dev/lupio/lupio_pic.cc
joel-porquet: I know that `{0}` works when it's within the same statement as the variable declaration itself (e.g., `int enable[10] = {0};`) but are you sure that it also works after the variable has been declared. Otherwise, another possibility would be to do the init as part of the initializer list? (disclaimer, I'm not a C++ guru)

src/dev/lupio/lupio_pic.cc
```diff
@@ -48,6 +49,11 @@ LupioPIC::LupioPIC(const Params &params) :
     nThread(params.num_threads),
     intType(params.int_type)
 {
+    memset(enable, 0, LUPIO_PIC_NCPU * sizeof(uint32_t));
+    memset(mask, 0, LUPIO_PIC_NCPU * sizeof(uint32_t));
+
+    // CPU0 receives all IRQ sources by default
+    enable[0] = ~0;
```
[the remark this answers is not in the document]
joel-porquet: All the bits of `enable[0]` should be set to `1`. We could also write `enable[0] = 0xFFFFFFFF`.

src/dev/lupio/lupio_pic.cc
powerjg: Ah, 0 is special. See https://en.cppreference.com/w/c/language/initialization#Zero_initialization

You can also write `enable = {}` to zero initialize as well.

powerjg (commented):
lgtm. Just one more small thing :)

joel-porquet (commented):
Almost there!

joel-porquet (commented):
Looks great to me :D

powerjg (approved):
Looks go to me too! Thanks for all of the hard work!

PR #245, opened 2022-08-04
Temporary PR for rebasing rvv stuff

hnpl:


src/arch/riscv/pcstate.hh
```diff
@@ -43,29 +43,39 @@
...
+
+class PCState : public GenericISA::UPCState<16>
 {
   private:
     bool _compressed = false;
     bool _rv32 = false;
 
   public:
-    using GenericISA::UPCState<4>::UPCState;
+    using GenericISA::UPCState<16>::UPCState;
```
takekoputa, line 70: This should not be changed to 16. UPCState<4> means the InstructionWidth of the micro op is 4-byte long.

src/arch/riscv/pcstate.hh
```diff
@@ -43,29 +43,39 @@
...
+#include "base/bitunion.hh"
 
 namespace gem5
 {
 
 namespace RiscvISA
 {
 
-class PCState : public GenericISA::UPCState<4>
+BitUnion32(VTYPE)
```
takekoputa, line 55: This should be elsewhere.

src/arch/riscv/isa/formats/standard.isa
```diff
@@ -198,7 +198,11 @@ def template BranchExecute {{
     %(class_name)s::branchTarget(const PCStateBase &branch_pc) const
     {
         auto &rpc = branch_pc.as<RiscvISA::PCState>();
-        return std::make_unique<PCState>(rpc.pc() + imm);
+        //return std::make_unique<PCState>(rpc.pc() + imm);
+        auto new_pc = std::make_unique<RiscvISA::PCState>(rpc.pc() + imm);
```
takekoputa, line 202: Not sure if this is necessary

PR #246, opened 2022-08-04
added the Simpoint Class and simpoints exit event support on set_se_workload and Simulator

studyztp:
Added a new Simpoint class under the src/python/gem5/components/processor/simpoint.py
It can take in either the simpoint and weight file paths or the simpoint and weight list.

Added the checkpoint generator and dump_reset generator.

The set_se_workload can take in Simpoint object and use the get function to pass the simpoint_start_insts to the core.

Added a schedule_simpoints function for scheduling simpoints in fs mode and schedule simpoint end if needed in Simulator.
Also added a schedule restore_simpoint_stop function in Simulator to schedule a stop when doing restore simpoint checkpoints.


powerjg:
Excited to check this out! Hopefully, I'll get a chance this afternoon :)

studyztp:
> 

Thank you!! No rush. :)

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,99 @@
+
```
powerjg, line 1: Needs a license

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,99 @@
...
+
+from m5.util import *
```
powerjg, line 3: It's better to avoid `import *` if possible.

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,99 @@
...
+        # with it in the future
+    ) -> None:
+        """
+        :param simpoint_interval: the length of each simpoint interval
+        :param simpoint_file_path: the path to the simpoint result file
+        generated by Simpoint3.2 or gem5
+        :param weight_file_path: the path to the weight result file generated
+        by Simpoint3.2 or gem5
+
+        :param simpoint_list: a list of simpoint starting instructions (sorted)
```
powerjg, line 32: Sorted how? What us the sorting?

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,99 @@
...
+        :param weight_list: a list of weights (sorted)
+
+        """
+
+        if simpoint_interval is None:
+            fatal("not passing in interval length")
```
powerjg, line 38: If it's required, then it shouldn't have a default None value :)

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,99 @@
...
+
+        self._simpoint_interval = simpoint_interval
+
+        if simpoint_file_path is None or weight_file_path is None:
+            if simpoint_list is None or weight_list is None:
+                # raise error
+                fatal("not passing in enough information")
```
powerjg, line 45: A more specific error would be good.

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,99 @@
...
+                    weight_file_path
+                    )
+
+    def get_weights_and_simpoints_from_file(
+        self,
+        simpoint_path:Path,
+        weight_path:Path,
+        )-> Tuple[List[int],List[int]]:
+        simpoint_file = open(simpoint_path)
+        weight_file = open(weight_path)
```
powerjg, line 66: I would suggest using `with open(filename) as simpoint_file:`

https://realpython.com/python-with-statement/#:~:text=The%20Python%20with%20statement%20creates,finally%20statement.

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,99 @@
...
+            line = weight_file.readline()
+            if not line:
+                fatal("not engough weights")
+            weight = float(line.split(" ",1)[0])
+            simpoint.append((interval,weight))
+        simpoint.sort(key=lambda obj: obj[0])
+        # use simpoint to sort
+        simpoint_start_insts = []
+        weight_list = []
+        for obj in simpoint:
```
powerjg, line 85:
```suggestion
        for start,weight in simpoint:
```

src/python/gem5/simulate/exit_event_generators.py
```diff
@@ -95,3 +98,29 @@ def default_workend_generator():
...
+        print("got to the simpoint_dump_reset_gnereator\n")
+        m5.stats.dump()
+        m5.stats.reset()
+        yield False
+
+def simpoint_save_checkpoint_generator(checkpoint_dir: Path):
+    while True:
+        # only print for testing purpose
+        print("got to the simpoint_save_checkpoint_generator\n")
+        m5.checkpoint(joinpath(checkpoint_dir,f"cpt.{str(m5.curTick())}"))
```
powerjg, line 125: We should think more about how to do this. I don't love that we're using the `m5` interface here. @BobbyRBruce, I think that we need to have a way to get to the simulator instance from these generators.

src/python/gem5/simulate/simulator.py
```diff
@@ -185,6 +190,18 @@ def unique_exit_event():
...
 
+    def schedule_simpoint(
+        self,
+        simpoint_end_inst: int = 0,
+        simpoint_start_insts: List[int] = []
+        ):
+        self._board.get_processor().get_cores()[0].\
+                        set_simpoint(simpoint_end_inst,simpoint_start_insts)
+
+    def schedule_simpoint_restore_stop(self, simpoint_interval: int):
```
powerjg, line 201: What is this function used for? Could you add pydoc for this and the function above?

powerjg (commented):
Generally, this looks great to me :)

src/cpu/base.cc
```diff
@@ -727,6 +727,28 @@ BaseCPU::traceFunctionsInternal(Addr pc)
...
+    {
+        for (size_t i = 0; i < starting_points.size(); ++i) {
+                scheduleInstStop(0, starting_points[i], cause1);
+        }
+    }
+    else {
+        scheduleInstStop(0, end_point-1, cause2);
+        // problem might occur when starting point = [0,1], because for inst 0,
+        // the event are scheduled at inst 1, and when inst 1 schedule an end
+        // point, the end point event will get scheduled at 1*interval length
```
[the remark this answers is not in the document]
studyztp: When I was doing testing, I encountered a conflict problem with non-KVM CPU when passing in simpoint_start_instruction = [0, 100000000] and schedule an end point when it received the SIMPOINT_BEGIN exit event. The end point event for simpoint 0 was raised after the SIMPOINT_BEGIN for simpoint 100000000 at instruction 100000000. I found that the SIMPOINT_BEGIN for simpoint 0 was scheduled at instruction 1, so when I scheduled an end point event for simpoint 0, it was schedule at instruction 1+end_point-1. In that case, there are two events scheduled at instruction 100000000, and it caused a conflict.

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -110,6 +110,26 @@ def get_mmu(self) -> BaseMMU:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def tesing(self) -> None:
+
+        raise NotImplementedError
+
+    @abstractmethod
+    def set_simpoint(
```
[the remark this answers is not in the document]
studyztp: The set_simpoint is for scheduling events after the simulation starts, and init_simpoint is for scheduling event before the simulation start. The use case for init_simpoint is the set_se_workload and restoring checkpoints. For example, when restoring a checkpoint, there isn't a beginning exit event for us to schedule the simpoint, so we need to schedule it before the simulation starts. Sorry about not having the comments and documentations, I'll add them and push them.

src/python/gem5/components/processors/simpoint.py
studyztp: The simpoint_list is a list of the simpoint interval numbers (ex. [3, 5, 15]). Sorry about the names, I'm bad at naming. They should be sorted in ascending order. This might be good for cases that need to schedule starting instructions one by one during the simulation. But now I think about it, you're right, they don't need to be sorted.

src/cpu/base.hh
```diff
@@ -178,6 +178,10 @@ class BaseCPU : public ClockedObject
      */
     virtual Port &getDataPort() = 0;
 
+    void scheduleSimpoint(
+      Counter end_point,
```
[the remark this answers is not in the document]
studyztp: The end_point is an instruction number. When user pass in an end_point, it will schedule a SIMPOINT_END exit event at the end_point+current local instructions.

src/python/gem5/components/processors/abstract_core.py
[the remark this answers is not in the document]
powerjg: This makes sense on the C++ side, but I wonder if it's absolutely needed on the python side. Could you detect that it's "init" and do the right thing.

src/python/gem5/components/processors/simpoint.py
powerjg: I see what you're saying now. Sorted based on the instruction number, not based on the weights or anything else.

src/python/gem5/components/processors/abstract_core.py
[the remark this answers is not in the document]
studyztp: Yes, I can add a boolean parameter to detect it. And I've been thinking should it have a boolean parameter to allow user to choose between scheduling SIMPOINT_BEGIN or SIMPOINT_END?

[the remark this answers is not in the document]
powerjg: The SIMPOINT_END is just SIMPOINT_BEGIN + interval_length, right?

I guess I'm a bit confused by that as well.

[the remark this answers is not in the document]
studyztp: I think of SIMPOINT_END as an exit event that's simpoints related but distinguish from the SIMPOINT_BEGIN. I can only think of use cases like doing switching CPUs in FS mode. Maybe I should get rid of it and change SIMPOINT_BEGIN to SIMPOINT_FOUND, so there is only one exit event for simpoints to make it less confusing.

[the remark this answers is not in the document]
powerjg: I definitely like SIMPOINT_BEGIN. However, maybe the "SIMPOINT_END" should just be "number of instructions reached" or whatever that exit event is...

src/cpu/base.hh
```diff
@@ -443,7 +443,28 @@ class BaseCPU : public ClockedObject
...
+    /**
+     * Schedule simpoint events using the scheduleInstStop function.
+     *
+     * This is used to raise a SIMPOINT_BEGIN exit event in the gem5 standard
+     * library.
+     *
+     * @param inst_starts A vector of number of instructions to start simpoints
+     */
+
+    void scheduleSimpoint(std::vector<Counter> inst_starts);
```
powerjg, line 457:
```suggestion
    void scheduleSimpointsInstStop(std::vector<Counter> inst_starts);
```

src/cpu/base.hh
```diff
@@ -443,7 +443,28 @@ class BaseCPU : public ClockedObject
...
+
+    /**
+     * Schedule an exit event when any threads in the core reach the max_insts
+     * instructions using the scheduleInstStop function.
+     *
+     * This is used to raise a MAX_INSTS exit event in thegem5 standard library
+     *
+     * @param max_insts Number of instructions into the future.
+     */
+    void scheduleOneMaxInsts(Counter max_insts);
```
powerjg, line 467:
```suggestion
    void scheduleInstStopAnyThread(Counter max_insts);
```

src/cpu/base.cc
```diff
@@ -727,6 +723,21 @@ BaseCPU::traceFunctionsInternal(Addr pc)
...
+    for (size_t i = 0; i < inst_starts.size(); ++i)
+        scheduleInstStop(0, inst_starts[i], cause);
+}
+
+void
+BaseCPU::scheduleOneMaxInsts(Counter max_insts)
+{
+    std::string cause = "a thread reached the max instruction count";
+    for (ThreadID tid = 0; tid < numThreads; ++tid)
+        scheduleInstStop(tid, max_insts, cause);
```
powerjg, line 739:
```suggestion
    for (ThreadID tid = 0; tid < numThreads; ++tid)
        scheduleInstStop(tid, max_insts, cause);
    }
```

configs/example/BBV/bbvprobe.py
```diff
@@ -0,0 +1,100 @@
...
+# download the binary from the gem5 Resources cloud bucket if it's not already
+# present.
+board.set_se_binary_workload(
+    # The `Resource` class reads the `resources.json` file from the gem5
+    # resources repository:
+    # https://gem5.googlesource.com/public/gem5-resource.
+    # Any resource specified in this file will be automatically retrieved.
+    # At the time of writing, this file is a WIP and does not contain all
+    # resources. Jira ticket: https://gem5.atlassian.net/browse/GEM5-1096
+    Resource("arm-hello64-static")
```
powerjg, line 88: It says x86 above :)

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,117 @@
+# Copyright (c) 2021 The Regents of the University of California
```
powerjg, line 1: I'm not sure this file makes sense in the "processors" package. Maybe in `util`? @BobbyRBruce, any opinion?

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -54,6 +60,10 @@ def set_se_binary_workload(
         :param binary: The resource encapsulating the binary to be run.
         :param exit_on_work_items: Whether the simulation should exit on work
         items. True by default.
+        :param stdin_file: The input file for the binary
+        :param arguments: The input arguments for the binary
+        :param simpoint: The SimPoint object that contains the list of simpoint
```
powerjg, line 65:
```suggestion
        :param simpoint: The SimPoint object that contains the list of simpoints
```

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -64,11 +74,21 @@ def set_se_binary_workload(
...
         process = Process()
-        process.cmd = [binary.get_local_path()]
+        process.executable = binary_path
+        process.cmd = [binary_path] + arguments
+        if stdin_file is not None:
+          process.input = stdin_file.get_local_path()
+
         self.get_processor().get_cores()[0].set_workload(process)
 
+        if simpoint is not None:
```
powerjg, line 88: We should probably add a warning that SimPoints only works with one core.

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -110,6 +110,32 @@ def get_mmu(self) -> BaseMMU:
...
+    ) -> None:
+        """ Schedule simpoint exit events for the core.
+
+        This is used to raise SIMPOINT_BEGIN exit events in the gem5 standard
+        library.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def set_one_max_insts(
```
powerjg, line 127: See name suggestion in `base.hh`

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -110,6 +110,32 @@ def get_mmu(self) -> BaseMMU:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def set_simpoint(
+        self,
+        inst_starts: List[int] = [],
+        init: bool = True
```
powerjg, line 117: Define these parameters :). Especially the `init` parameter.

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,117 @@
...
+    ) -> None:
+        """
+        :param simpoint_interval: the length of each simpoint interval
+        :param simpoint_file_path: the path to the simpoint result file
+        generated by Simpoint3.2 or gem5
+        :param weight_file_path: the path to the weight result file generated
+        by Simpoint3.2 or gem5
+
+        :param simpoint_list: a list of simpoint starting instructions
+        :param weight_list: a list of weights
```
powerjg, line 57: Make it clear that you need either the path or the list in this help documentation.

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,117 @@
...
+        return simpoint_start_insts, weight_list
+
+    def get_simpoint_start_insts(self)->List[int]:
+        return self._simpoint_start_insts
+
+    def get_weight_list(self)->List[float]:
+        return self._weight_list
+
+    def get_simpoint_interval(self)->int:
+        return self._simpoint_interval
```
powerjg, line 117: empty newlines at the end of the file are nice

src/python/gem5/simulate/exit_event_generators.py
```diff
@@ -28,6 +28,9 @@
 from ..components.processors.abstract_processor import AbstractProcessor
 from ..components.processors.switchable_processor import SwitchableProcessor
 from m5.util import warn
+from os.path import join as joinpath
+from pathlib import Path
```
powerjg, line 32: Do we need both of these?

If using pathlib, I think you can use this instead: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.joinpath

src/python/gem5/simulate/exit_event_generators.py
```diff
@@ -28,6 +28,9 @@
...
+import os
```
powerjg, line 33: Is this used?

src/python/gem5/simulate/simulator.py
```diff
@@ -185,6 +190,38 @@ def unique_exit_event():
...
+        )-> None:
+        """
+        Schedule SIMPOINT_BEGIN exit events
+
+        :param simpoint_start_insts: a list of number of instructions
+        indicating the starting point of the simpoints
+        :param schedule_at_init: if it is True, schedule the events in the init
+        stage of the core, else, schedule the events during the simulation
+        """
+        self._board.get_processor().get_cores()[0].\
```
powerjg, line 206: Warn if more than 1 core?

src/python/gem5/simulate/simulator.py
```diff
@@ -185,6 +190,38 @@ def unique_exit_event():
...
+        """
+        Schedule a MAX_INSTS exit event when any thread in the current core
+        reaches the given number of instructions
+
+        :param insts: a number of instructions
+        :param schedule_at_init: if it is True, schedule the event in the init
+        stage of the core, else, schedule the event during the simulation
+        """
+        self._board.get_processor().get_cores()[0].\
+                        set_one_max_insts(inst,schedule_at_init)
```
powerjg, line 223:
```suggestion
                        set_one_max_insts(inst, schedule_at_init)
```

src/python/gem5/simulate/simulator.py
```diff
@@ -185,6 +190,38 @@ def unique_exit_event():
...
+
+        :param simpoint_start_insts: a list of number of instructions
+        indicating the starting point of the simpoints
+        :param schedule_at_init: if it is True, schedule the events in the init
+        stage of the core, else, schedule the events during the simulation
+        """
+        self._board.get_processor().get_cores()[0].\
+                        set_simpoint(simpoint_start_insts,schedule_at_init)
+
+    def schedule_max_insts(
```
powerjg, line 209: For this function, we definitely don't want to assume just 1 core. I'm not certain I know *what* we want to do, though.

I think maybe this should be a board function / processor function.

powerjg (commented):
Can you add an example runscript that uses simpoints? I think this looks good, but I'd like to see the external user interface.

studyztp:
> Can you add an example runscript that uses simpoints? I think this looks good, but I'd like to see the external user interface.

Yes, I'll add an example runscript. I'll push it later with the fixed files and ping you and Bobby. Thank you for reviewing :)

src/python/gem5/simulate/simulator.py
studyztp: Maybe it should be a processor function because it's more related to a processor than to a board..?

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+)
+
+from gem5.components.processors.simpoint import SimPoint
+from gem5.simulate.exit_event_generators import (
+    simpoint_save_checkpoint_generator,
+)
+
+requires(isa_required=ISA.X86)
+
+cache_hierarchy = PrivateL1PrivateL2CacheHierarchy(
```
powerjg, line 23: I suggest using `NoCache` when taking checkpoints for SimPoints (at least in our examples). When restoring from a checkpoint, there's no cache state saved, so you can completely change the cache hierarchy. By using `NoCache` to take the checkpoints, we'll get slightly better performance when running in atomic mode, and it won't put any restrictions on what people can do *with* the checkpoints. 

As an aside, when we finalize these scripts as examples, we should write things like the above in comments so everyone can learn.

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+    l1d_size = "32kB",
+    l1i_size="32kB",
+    l2_size="256kB",
+)
+
+memory = DualChannelDDR4_2400(size = "3GB")
```
powerjg, line 29: Similarly, we can use `SimpleMemory` to (possibly) improve performance.. Though, TBH, I'm not sure this makes any difference.

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+
+processor = SimpleProcessor(
+    cpu_type=CPUTypes.ATOMIC,
+    isa=ISA.X86,
+    num_cores=1,
```
powerjg, line 34: Perfect. I would note that SimPoints requires 1 core.

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
```diff
@@ -0,0 +1,70 @@
...
+dir.mkdir(exist_ok=True)
+
+simulator = Simulator(
+    board=board,
+    on_exit_event={
+        # using the SimPoints event generator in the standard library to take
+        # checkpoints
+        ExitEvent.SIMPOINT_BEGIN: simpoint_save_checkpoint_generator(dir)
+    }
+)
```
powerjg, line 68: I LOVE THIS! I'm sooo happy seeing this code!!

configs/example/gem5_library/checkpoints/simpoints-se-restore.py
```diff
@@ -0,0 +1,70 @@
...
+from gem5.components.memory import DualChannelDDR4_2400
+from gem5.components.processors.simple_processor import SimpleProcessor
+from gem5.components.processors.cpu_types import CPUTypes
+from gem5.isas import ISA
+from gem5.resources.resource import Resource
+
+from gem5.components.processors.simpoint import SimPoint
+from pathlib import Path
+
+cache_hierarchy = PrivateL1PrivateL2CacheHierarchy(
```
powerjg, line 18: *Here* is a good place to use the more complex hierarchy. This is great. Make a note in a comment that you can change the hierarchy :)

configs/example/gem5_library/checkpoints/simpoints-se-restore.py
```diff
@@ -0,0 +1,70 @@
...
+    l1d_size = "32kB",
+    l1i_size="32kB",
+    l2_size="256kB",
+)
+
+requires(isa_required=ISA.X86)
+
+memory = DualChannelDDR4_2400(size = "3GB")
```
powerjg, line 26: Same as above. This is great. Let's add a note about why it's OK to change this.

configs/example/gem5_library/checkpoints/simpoints-se-restore.py
```diff
@@ -0,0 +1,70 @@
...
+
+simulator = Simulator(
+    board=board,
+    checkpoint_path=dir,
+    on_exit_event={
+        ExitEvent.MAX_INSTS: exit()
+    }
+)
+# schedule a MAX_INSTS exit event before the simulation begins
+simulator.schedule_max_insts(simpoint.get_simpoint_interval(), True)
```
powerjg, line 69: Can you add a "warmup" period here. I'd like to see how it works to warmup for *x* instructions then run the simpoint for *interval - x* instructions.

Hmm, now that I write that out, maybe it's not quite right. Can you go back to the SimPoint paper and see how they talked about warmup.

src/cpu/base.cc
```diff
@@ -727,6 +723,27 @@ BaseCPU::traceFunctionsInternal(Addr pc)
...
 }
 
+void
+BaseCPU::scheduleSimpointsInstStop(std::vector<Counter> inst_starts)
+{
+    std::string cause = "simpoint starting point found";
+    for (size_t i = 0; i < inst_starts.size(); ++i){
+
+        scheduleInstStop(0, inst_starts[i], cause);
+
```
powerjg, line 733: I know I'm being nitpicky here but... can you remove the extra newlines? Also, there should be a space toward the end of the `for` line (`){` -> `) {`)

src/cpu/base.cc
```diff
@@ -727,6 +723,27 @@ BaseCPU::traceFunctionsInternal(Addr pc)
...
+    }
+}
+
+void
+BaseCPU::scheduleInstStopAnyThread(Counter max_insts)
+{
+    std::string cause = "a thread reached the max instruction count";
+    for (ThreadID tid = 0; tid < numThreads; ++tid){
+
```
powerjg, line 742: Same as above.

src/python/gem5/simulate/simulator.py
powerjg: I'll drop this comment. I'm trying to look for too much "perfection" and I'm not really sure there's a better way.

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
studyztp: XD

configs/example/gem5_library/checkpoints/simpoints-se-restore.py
studyztp: They mentioned warmup in section 5 and 6. I believe they did the warmup by doing the fast-forwarding: "all
fast forwarding is done so that all of the architecture structures
are completely warmed up when starting simulation (no
cold-start effect)." Later in section 6, the related work section, they mentioned using the entire fast-forward interval for warmup should be the minimum set of fast-forward transactions needed to get to the accurately produce state. They also mentioned checkpoints can either store the architectural state or warm up in a similar manner to the above method.

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
studyztp: I'll change to NoCache and add the comments. It's a very useful information. Thank you!

configs/example/gem5_library/checkpoints/simpoints-se-restore.py
studyztp: I'm not familiar with warmup but I'll try taking the checkpoint x instructions earlier than the SimPoint and execute the x instructions to perform warmup before taking statistics for the SimPoint. In this case, maybe the number of instructions for warmup should be stored in the SimPoint class too. Please let me know if I misunderstand the warmup process.

powerjg: Yeah, that sounds great. I'm good with adding a warmup instructions parameter to the simpoint object.

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
studyztp: When I tried to use SimpleMemory, there is an error caused by it not being inherited from the AbstractMemorySystem. It doesn't have the 'get_size' function the stdlib board needs during the its init stage. So I used the SingleChannelDDR3_1600 when taking checkpoints, and changed to DualChannelDDR4_2400 when restoring from a checkpoint. Does it sounds correct?

studyztp: During the testing, I also encountered an error when I had two different sizes setup for the memory when taking checkpoints and restoring from one. Does it mean the memory size has to be the same?

powerjg: Yes and yes 🙂

src/python/gem5/components/processors/simpoint.py
BobbyRBruce: Yeah. I don't think this file has a logical home, but i think `util` makes the most sense here.


Also, the copyright should be 2022, not 2021.

src/python/gem5/components/processors/simpoint.py
```diff
@@ -0,0 +1,171 @@
...
+        :param weight_list: a list of SimPoints weights
+        :param warmup_interval: a number of instructions for warming up before
+        restoring a SimPoints checkpoint
+
+        usage note
+        -----------
+        Need to pass in the paths or the lists for the SimPoints and their
+        weights. If the paths are passed in, no actions will be done to the
+        list.
+
```
BobbyRBruce, line 67: Perhaps given this usage there should be two `def __init__` functions: one which accepts paths, and another that accepts lists. Is this possible to do? It'd make it much clearer to the user how this is used.

src/python/gem5/simulate/simulator.py
```diff
@@ -185,6 +188,42 @@ def unique_exit_event():
...
+
+        **Warning:** SimPoints only work with one core
+
+        :param simpoint_start_insts: a list of number of instructions
+        indicating the starting point of the simpoints
+        :param schedule_at_init: if it is True, schedule the events in the init
+        stage of the core, else, schedule the events during the simulation
+        """
+        if self._board.get_processor().get_num_cores() > 1:
+            warn("SimPoints only work with one core")
```
BobbyRBruce, line 207: Should this just be a warning? Maybe it should be a proper exception? A warning won't stop a user doing this, but maybe we should.

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -66,7 +66,6 @@ def connect_icache(self, port: Port) -> None:
         """
         This function should connect the response port from the instruction
         cache to the right request port on the core.
-
```
BobbyRBruce, line 69: I think there's some weird formatting here where these new lines have been removed. Can you re-add them?

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -116,7 +112,38 @@ def connect_interrupt(
     @abstractmethod
     def get_mmu(self) -> BaseMMU:
         """Return the MMU for this core.
-
         This is used in the board to setup system-specific MMU settings.
         """
         raise NotImplementedError
+
+    @abstractmethod
+    def set_simpoint(
```
BobbyRBruce, line 120: Good job working with my latest changes!

Everything here looks good but I'd add a helpful message to the `NotImplementedError`, something like:

```
raise NotImplementedError("This core type does not support simpoints")
```

I'm thinking about the case where we have cores that are not BaseCPUCores. If someone tries to set and use simpoints a vanilla `NotImplementedError` might be confusing.

configs/example/gem5_library/checkpoints/simpoints-se-restore.py
```diff
@@ -0,0 +1,89 @@
...
+from gem5.components.memory import DualChannelDDR4_2400
+from gem5.components.processors.simple_processor import SimpleProcessor
+from gem5.components.processors.cpu_types import CPUTypes
+from gem5.isas import ISA
+from gem5.resources.resource import Resource
+
+from gem5.components.processors.simpoint import SimPoint
+from pathlib import Path
+from m5.stats import reset, dump
+
```
BobbyRBruce, line 19: Maybe just add a bit of documentation here explaining that this should be run after "simpoints-se-checkpoint.py".

configs/example/gem5_library/checkpoints/simpoints-se-checkpoint.py
```diff
@@ -0,0 +1,72 @@
+from gem5.simulate.exit_event import ExitEvent
```
BobbyRBruce, line 1: This file is lacking a copyright message.

configs/example/gem5_library/checkpoints/simpoints-se-restore.py
```diff
@@ -0,0 +1,89 @@
+import imp
```
BobbyRBruce, line 1: This file is lacking a copyright message.

BobbyRBruce (changes requested):
Very well done! I have a few questions/comments, but for the most part I agree with most of these changes. It's good work!

src/python/gem5/components/processors/simpoint.py
powerjg: I'll push back here. Python doesn't support differentiating functions on input types (it's duck typed!). So, I think this is actually a great way to do it.

You *could* have two function `_init_lists()` and `_init_files()` which the `__init__` calls, but I'm not sure that's any better.

src/python/gem5/simulate/simulator.py
powerjg: Good question. It won't *break* anything to have more than 1 core. I think a warning is the right level.

src/python/gem5/components/processors/simpoint.py
BobbyRBruce: Fair, was just wondering if there was a nice way to do it. If not then go with this.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -82,5 +89,12 @@ def set_se_binary_workload(
 
         self.get_processor().get_cores()[0].set_workload(process)
 
+        if simpoint is not None:
+          if self.get_processor().get_num_cores() > 1:
+            warn("SimPoints only works with one core")
```
powerjg, line 94: Needs two more spaces (run black?)

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -120,3 +120,35 @@ def get_mmu(self) -> BaseMMU:
...
+        self, inst_starts: List[int] = [], init: bool = True
```
powerjg, line 126: Why do these parameters have default values? Are neither of them required? `inst_starts` seems required at a minimum.

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -120,3 +120,35 @@ def get_mmu(self) -> BaseMMU:
...
+        :param inst_starts: a list of SimPoints starting instructions
+        :param init: if it is True, the starting instructions will be scheduled
+        at the init stage of the core, else, the starting insructions will be
+        scheduled during the simulation
+        """
+        raise NotImplementedError("This core type does not support simpoints")
+
+    @abstractmethod
+    def set_inst_stop_any_thread(
+        self, inst: int = 0, init: bool = True
```
powerjg, line 142: Why do these parameters have default values? Are neither of them required? `inst` seems required at a minimum as it doesn't make sense to stop at instruction 0, right?

src/python/gem5/components/processors/base_cpu_core.py
```diff
@@ -151,3 +151,21 @@ def connect_interrupt(
     @overrides(AbstractCore)
     def get_mmu(self) -> BaseMMU:
         return self.core.mmu
+
+    @overrides(AbstractCore)
+    def set_simpoint(
+        self, inst_starts: List[int] = [], init: bool = True
```
powerjg, line 157: Ditto about default values.

src/python/gem5/components/processors/base_cpu_core.py
```diff
@@ -151,3 +151,21 @@ def connect_interrupt(
...
+    ) -> None:
+        if init:
+            self.core.simpoint_start_insts = inst_starts
+        else:
+            self.core.scheduleSimpointsInstStop(inst_starts)
+
+    @overrides(AbstractCore)
+    def set_inst_stop_any_thread(
+        self, inst: int = 0, init: bool = True
```
powerjg, line 166: Same question on default values.

src/python/gem5/simulate/exit_event_generators.py
```diff
@@ -103,3 +104,33 @@ def default_workend_generator():
...
+        "A default generator for SimPoints. It will do nothing.")
+    while True:
+        yield False
+
+def simpoint_dump_reset_generator():
+    while True:
+        """
+            A generator for SimPoints. It will reset the simulation statistics
+            and then dump simulation statistics.
+        """
```
powerjg, line 124:
```suggestion
        """
            A generator for SimPoints. It will reset the simulation statistics
            and then dump simulation statistics.
        """
    while True:
```

src/python/gem5/simulate/exit_event_generators.py
```diff
@@ -103,3 +104,33 @@ def default_workend_generator():
...
+        m5.stats.dump()
+        m5.stats.reset()
+        yield False
+
+def simpoint_save_checkpoint_generator(checkpoint_dir: Path):
+    while True:
+        """
+            A generator for SimPoints. It will take a checkpoint with the input
+            path and the current simulation Ticks.
+        """
```
powerjg, line 134:
```suggestion
        """
            A generator for SimPoints. It will take a checkpoint with the input
            path and the current simulation Ticks.
        """
    while True:
```

src/python/gem5/simulate/exit_event_generators.py
```diff
@@ -103,3 +104,33 @@ def default_workend_generator():
...
+    """
+    A default generator for SimPoints. It will do nothing.
+    """
+    defaultBehaviorWarning(
+        "default_simpoint_generator",
+        "A default generator for SimPoints. It will do nothing.")
+    while True:
+        yield False
+
+def simpoint_dump_reset_generator():
```
powerjg, line 119: Does this have to be "simpoint" specific, or is this just a dump/reset generator?

src/python/gem5/simulate/exit_event_generators.py
```diff
@@ -103,3 +104,33 @@ def default_workend_generator():
...
+    while True:
+        """
+            A generator for SimPoints. It will reset the simulation statistics
+            and then dump simulation statistics.
+        """
+        m5.stats.dump()
+        m5.stats.reset()
+        yield False
+
+def simpoint_save_checkpoint_generator(checkpoint_dir: Path):
```
powerjg, line 129: Ditto, is this simpoint-specific?

src/python/gem5/utils/simpoint.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(
+        self,
+        simpoint_interval: int,
+        simpoint_file_path: Path = None,
+        weight_file_path: Path = None,
+        simpoint_list: List[int] = None,
+        weight_list: List[int] = None,
+        warmup_interval: int = 0
+        # bbv_file_path: Path = None, will take in bbv and generate simpoints
+        # with it in the future
```
powerjg, line 48: Do we really need to commit to this? I suggest removing this comment

src/python/gem5/utils/simpoint.py
```diff
@@ -0,0 +1,171 @@
...
+            ) = self.get_weights_and_simpoints_from_file(
+                simpoint_file_path, weight_file_path
+            )
+
+        if warmup_interval != 0:
+            self._warmup_list = self.set_warmup_intervals(warmup_interval)
+        else:
+            self._warmup_list = [
+                0 for _ in range(len(self._simpoint_start_insts))
+            ]
```
powerjg, line 100:
```suggestion
            self._warmup_list = [0] * len(self._simpoint_start_insts)
```

powerjg (commented):
I think the only thing left to do is to split this into specific commits. Other than the small comments that remain, we're almost ready to submit to gerrit!

studyztp:
@powerjg How about splitting them into an exit_event_generator commit, a simpoints module commit, a base cpu simpoints related scheduling function commit, and a stdlib simpoints support commit(the simulator, exit event, and se_set_workload)?

PR #247, opened 2022-08-15
RISC-V Board

kunpai:
As a component of the stdlib, the cache, core, processor etc. inherit from various components already existing in the standard library, to model the HiFive Unmatched as closely as possible.


src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -23,52 +23,49 @@
 # THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
```
powerjg, line 26: I suggest not committing whitespace-only changes unless you're updating the style to be more compliant.

src/python/gem5/prebuilt/hifiveunmatched/README.md
```diff
@@ -0,0 +1,52 @@
+# HiFive GEM5 Board Specs and Usage
+
+## Specs for the HiFive Unmatched
+
+### Cache Hierarchy
+- L1D Size : 32kB
+- L1D Associativity : 8
+- L1I Size : 32kB
+- L1I Associativity : 4
+- L2 Size : 2MB
```
powerjg, line 10: Note (in theory) this is actually configurable. We may want to make this a parameter of the board.

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) 2022 The Regents of the University of California
```
powerjg, line 1: The "runscript" shouldn't be in `src/`. It should be in `configs/examples` or `configs/known_good`.

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
```diff
@@ -0,0 +1,59 @@
...
+from gem5.simulate.simulator import Simulator
+from hifive_board import HiFiveUnmatchedBoard
+from typing import List
+
+# collect optional CLI arg for RISCV binary to run
+parser = argparse.ArgumentParser(description="Binary to run on system")
+parser.add_argument(
+    "--riscv_binary", type=str, help="The RISCV binary to execute on the CPU",
+    default="riscv-hello",
+)
```
powerjg, line 38: A required argument shouldn't be `--<something>`. Also, an argument that the user *must* consider (e.g., what workload they want to run) shouldn't have a default.

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
```diff
@@ -0,0 +1,59 @@
...
+
+# run default binary if no binary provided, otherwise run CLI provided resource
+if args.riscv_binary == "riscv-hello":
+    board.set_se_binary_workload(
+        Resource(args.riscv_binary)
+    )
+else:
+    board.set_se_binary_workload(
+        CustomResource(args.riscv_binary),
+        arguments=[args.argv]
```
powerjg, line 55: You can remove the `if`/`else` if you use `args.argv.split()`

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+simulator = Simulator(board=board)
+simulator.run()
```
powerjg, line 59: One newline at the end of the file is nice.

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
```diff
@@ -0,0 +1,59 @@
...
+    "--riscv_binary", type=str, help="The RISCV binary to execute on the CPU",
+    default="riscv-hello",
+)
+parser.add_argument(
+    "--argv", type=str, help="CLI argument to the binary",
+    default=""
+)
+args = parser.parse_args()
+
+board = HiFiveUnmatchedBoard()
```
powerjg, line 45: I love how simple this is!

src/python/gem5/prebuilt/hifiveunmatched/hifive_board.py
```diff
@@ -0,0 +1,56 @@
...
+from hifive_proc import U74Processor
+from m5.objects import AddrRange
+
+class HiFiveUnmatchedBoard(SimpleBoard):
+    def __init__(self) -> None:
+        requires(isa_required=ISA.RISCV)
+
+        cache_hierarchy = HiFiveCacheHierarchy(
+            l1d_size="32kB", l1i_size="32kB", l2_size="2MB"
+        )
```
powerjg, line 43: Perfect!

src/python/gem5/prebuilt/hifiveunmatched/hifive_board.py
```diff
@@ -0,0 +1,56 @@
...
+
+        memory = ChanneledMemory(DDR4_2400_8x8, 1, 64, "16GB")
+        memory.set_memory_range(
+            [AddrRange(start=0x80000000, size=memory.get_size())])
+
+        processor = U74Processor()
```
powerjg, line 49: Wonderful!

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
```diff
@@ -0,0 +1,185 @@
...
+        # ITLB Page walk caches
+        self.iptw_caches = [
+            MMUCache(size="4KiB")
+            for _ in range(board.get_processor().get_num_cores())
+        ]
+        # DTLB Page walk caches
+        self.dptw_caches = [
+            MMUCache(size="4KiB")
+            for _ in range(board.get_processor().get_num_cores())
+        ]
```
powerjg, line 142: Out of curiosity, are these listed in the spec?

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
```diff
@@ -0,0 +1,185 @@
...
+from gem5.utils.override import *
+
+
+class HiFiveCacheHierarchy(
+    AbstractClassicCacheHierarchy, AbstractTwoLevelCacheHierarchy
+):
+    """
+    A cache setup where each core has a private L1 Data and Instruction Cache,
+    and a private L2 cache.
+    """
```
powerjg, line 53: Point to where the information is for this hierarchy in the spec.

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
```diff
@@ -0,0 +1,185 @@
...
+        membus.badaddr_responder = BadAddr()
+        membus.default = membus.badaddr_responder.pio
+        return membus
+
+    def __init__(
+        self,
+        l1d_size: str,
+        l1i_size: str,
+        l2_size: str,
+        membus: BaseXBar = _get_default_membus.__func__(),
```
powerjg, line 74: Are these parameters needed?

I would argue that there should be just one parameter, which is the programmable size of the L2 (number of ways??? I can't remember how it's programmed specifically).

src/python/gem5/prebuilt/hifiveunmatched/hifive_core.py
```diff
@@ -0,0 +1,154 @@
...
+    def connect_interrupt(
+        self, interrupt_requestor: Optional[Port] = None,
+        interrupt_responce: Optional[Port] = None
+    ) -> None:
+        # TODO: This model assumes that we will only create an interrupt
+        # controller as we require it. Not sure how true this is in all cases.
+        self.core.createInterruptController()
+    @overrides(AbstractCore)
+    def get_mmu(self) -> BaseMMU:
+        return self.core.mmu
```
powerjg, line 154: I suggest adding some whitespace :)

src/python/gem5/prebuilt/hifiveunmatched/hifive_proc.py
```diff
@@ -0,0 +1,64 @@
...
+    ) -> None:
+        self._cpu_type = CPUTypes.MINOR
+        super().__init__(
+            cores=self._create_cores()
+        )
+
+    def _create_cores(
+        self,
+    ):
+        return U74Core()
```
powerjg, line 58: Shouldn't this be `[U74Core() for i in range(4)]`?

src/python/gem5/prebuilt/hifiveunmatched/hifive_proc.py
```diff
@@ -0,0 +1,64 @@
...
+        self,
+    ) -> None:
+        self._cpu_type = CPUTypes.MINOR
+        super().__init__(
+            cores=self._create_cores()
+        )
+
+    def _create_cores(
+        self,
+    ):
```
powerjg, line 57: No need for the newlines here.

powerjg (commented):
This looks *really* great! I'm excited to see the data!

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
jasrangi: The location is now configs/example/gem5_library/

src/python/gem5/prebuilt/hifiveunmatched/hifive_proc.py
kunpai: Hi Jason! Thanks for the feedback. I believe the MinorCPU in stdlib only runs one core. When we tried to set the number of cores as 4, it panicked. @jasrangi can you confirm that I recall correctly?

powerjg: You'll need to set them all up correctly 😆.

I guess 1 core is OK for now. But we need to be able to support the same configuration at some point.

kunpai: Oh I see, yeah I agree with you. We might need to set it up later. For now, even when we run with ./perf we taskset the workload so that there is no migration of CPU cores

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
kunpai: Yeah, I might be mistaken in interpreting it, but on page 55 (section 4.7.7) of the datasheet, there is a mention of the level 2 TLB pagesize being 4 KiB. On writing this, I realize the importance of citing the page number and section from where we picked up the data :)

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
jasrangi: Removed the default and set it to "riscv_binary" instead of "--riscv_binary".

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
jasrangi: The `if`/`else` exists to use Resource() when riscv-hello must be run, and use CustomResource() otherwise. Is there a better way to do that?
Also I pushed a change that uses split() to allow for multiple arguments in case future benchmarks require more than one.

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
powerjg: A two level TLB and a PWC are two different things. The current implementation is good with me, but we should make a note of the discrepancy.

src/python/gem5/prebuilt/hifiveunmatched/hifive-run.py
powerjg: I think that it should be a CustomResource in all cases. You the argument is the path to a binary, not the name of a resource.

It would be confusing if you could use the same argument to specify either a resource (of which there is only one? What happens when there's more?) or a path.

jasrangi: Ok that makes sense. It is now using CustomResource() only.

kunpai:
src/python/gem5/components/boards/se_binary_workload.py is a file we had Zhantong provide to us based on her work, how do I remove it from this PR?

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
```diff
@@ -0,0 +1,175 @@
...
+        A method used to obtain the default memory bus of 64 bit in width for
+        the PrivateL1PrivateL2 CacheHierarchy.
+        :returns: The default memory bus for the PrivateL1PrivateL2
+        CacheHierarchy.
+        :rtype: SystemXBar
+        """
+        membus = SystemXBar(width=64)
+        membus.badaddr_responder = BadAddr()
+        membus.default = membus.badaddr_responder.pio
+        return membus
```
powerjg, line 67: I don't think this is necessary. We don't need this "option" in the HiFive-specific board.

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __init__(
+        self,
+        l2_size: str,
+        membus: BaseXBar = _get_default_membus.__func__(),
```
powerjg, line 72: Ditto. I don't think this parameter is needed.

src/python/gem5/prebuilt/hifiveunmatched/hifive_cache.py
```diff
@@ -0,0 +1,175 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from gem5.components.cachehierarchies.abstract_cache_hierarchy import (
+    AbstractCacheHierarchy,
+)
+from gem5.components.cachehierarchies.classic.abstract_classic_cache_hierarchy import (
+    AbstractClassicCacheHierarchy,
+)
+from gem5.components.cachehierarchies.abstract_two_level_cache_hierarchy import (
+    AbstractTwoLevelCacheHierarchy,
```
powerjg, line 34: @BobbyRBruce, can you think about whether it makes sense to inherit from this class?

powerjg (commented):
Overall, it looks good.

One big thing that's missing is support for full system mode. Let's start focusing on that so we can get some more realistic results.

src/python/gem5/prebuilt/hifiveunmatched/hifive_board.py
```diff
@@ -0,0 +1,579 @@
...
+            self.platform.plic.n_contexts = self.processor.get_num_cores() * 2
+            self.platform.attachPlic()
+            self.platform.clint.num_threads = self.processor.get_num_cores()
+
+            # Add the RTC
+            """
+            The frequency of the RTC for the system is set to 1MHz. 
+            Details can be found on page 77, section 7.1 of the datasheet.
+
+            """
```
powerjg, line 151: `"""` should only be used for pydoc statements, not for general comments

src/cpu/minor/execute.cc
```diff
@@ -884,6 +884,47 @@ Execute::doInstCommitAccounting(MinorDynInstPtr inst)
     cpu.stats.committedInstType[inst->id.threadId]
                                [inst->staticInst->opClass()]++;
 
+    /** Add a count for every control instruction */
+
+    if (inst->staticInst->isControl())
+    {
+        if (inst->staticInst->isReturn())
+        {
```
powerjg:
```suggestion
    if (inst->staticInst->isControl()) {
        if (inst->staticInst->isReturn()) {
```
brackets should be on the same line.

src/cpu/minor/stats.cc
```diff
@@ -79,6 +95,10 @@ MinorStats::MinorStats(BaseCPU *base_cpu)
         .init(base_cpu->numThreads, enums::Num_OpClass)
         .flags(statistics::total | statistics::pdf | statistics::dist);
     committedInstType.ysubnames(enums::OpClassStrings);
+
+    commitedControl
+        .init(base_cpu->numThreads, enums::ControlInst::NumControlInstTypes);
```
powerjg: Can you use `StaticInstFlags` instead?

src/python/gem5/prebuilt/hifiveunmatched/hifive_board.py
```diff
@@ -0,0 +1,577 @@
...
+            self=self,
+            kernel=kernel,
+            disk_image=disk_image,
+            bootloader=bootloader,
+            readfile=readfile,
+            readfile_contents=readfile_contents,
+            kernel_args=kernel_args,
+            exit_on_work_items=exit_on_work_items,
+        )
+
```
powerjg, line 577: Something is wrong with this whitespace :)

src/cpu/minor/stats.cc
kunpai: Yeah, sure, but these functions are predefined and return the StaticInstFlags I believe. Just so I understand gem5 better, why would we not use this function?

powerjg: I was suggesting this to avoid creating a new enum.

jasrangi: Enum StaticInstFlags.Flags of build/RISCV/enums/StaticInstFlags.hh has 36 different flags. From my understanding, if we use this our statistics vector will be of length 36 (currently it is 7). Would it be better to use those flags and extend the statistics to track all 36 StaticInst types?

powerjg: I think it is better to extend the stat vector to be length 36, but only update the stats `if (isControl())`. You can set the option on the statistic so that it only displays if the value is non-zero to hide all of the others.

kunpai: So use .flags(statistics::dist) I believe, right?

powerjg: I don't think so. See https://www.gem5.org/documentation/general_docs/statistics/ you can use the `nozero` flag.

kunpai: Oh yeah you're right, I was reading the wrong comment in src/base/stats/info.hh lol

jasrangi: Removed the old enum and string map and used StaticInstFlags. Commit [315ae16](https://github.com/darchr/gem5/pull/247/commits/315ae16cbfad64f56622f14cb9437ec2f958fbe7).

PR #249, opened 2022-09-16
Looppoint

studyztp:
This implementation allow gem5 to take in and keep track on the input list of target PCs and their target counts. It can raise an exit event when the PC counts of the target PC reaches the target count during the simulation. With the exit events, we can take and restore checkpoints to do detail analysis on the selected regions. 
The implementation of this is mainly based on the Probe object in gem5. The Probe object is designed for custom tracking. In the BaseCPU, there is a ProbePoint that keeps track of committed instruction PC. This ProbePoint will notify its ProbeListener every time a core finishes committing. The LoopPoint is a ProbeListnerObject that’s receiving and handling the notification from each core. Because of the design of the Probe, one ProbeListener can only listen to one SimObject’s ProbePoints. Therefore, there is a LoopPointManager SimObject to handle the PC counting for multithread.


src/cpu/BaseCPU.py
```diff
@@ -315,3 +317,10 @@ def __init__(self, **kwargs):
                 self._uncached_interrupt_request_ports
                 + ["interrupts[0].int_requestor"]
             )
+
+    def addLoopPointProbe(self, targetpc, manager):
```
powerjg: Let's put this in the stdlib instead of in the SimObject description file. I'd like to start migrating things out of this file into the stdlib more generally :)

src/cpu/probes/LoopPoint.py
```diff
@@ -0,0 +1,13 @@
+from m5.params import *
```
powerjg, line 27: Needs copyright

src/cpu/probes/LoopPoint.py
```diff
@@ -0,0 +1,13 @@
...
+
+class LoopPoint(ProbeListenerObject):
+
+    type = "LoopPoint"
+    cxx_header = "cpu/probes/looppoint.hh"
+    cxx_class = "gem5::LoopPoint"
+
+    target_pc = VectorParam.UInt64(0, "the target PC")
+    core = Param.BaseCPU(NULL, "the connected cpu")
+    LPmanager = Param.LoopPointManager(NULL, "the looppoint manager")
```
powerjg:
```suggestion
    manager = Param.LoopPointManager(NULL, "the looppoint manager")
```

I'd just call this the `manager`. This way you don't have to violate the style with a capital first letter.

Also, what does it mean to be NULL here? Is that valid? If not, then we should just have no default.

src/cpu/probes/LoopPoint.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class LoopPoint(ProbeListenerObject):
+
+    type = "LoopPoint"
+    cxx_header = "cpu/probes/looppoint.hh"
+    cxx_class = "gem5::LoopPoint"
+
+    target_pc = VectorParam.UInt64(0, "the target PC")
+    core = Param.BaseCPU(NULL, "the connected cpu")
```
powerjg: Also, what does it mean to be NULL here? Is that valid? If not, then we should just have no default.

src/cpu/probes/LoopPoint.py
```diff
@@ -0,0 +1,13 @@
...
+from m5.objects.Probe import ProbeListenerObject
+
+
+class LoopPoint(ProbeListenerObject):
+
+    type = "LoopPoint"
+    cxx_header = "cpu/probes/looppoint.hh"
+    cxx_class = "gem5::LoopPoint"
+
+    target_pc = VectorParam.UInt64(0, "the target PC")
```
powerjg: What does it mean to be 0 here? Is that valid? If not, then we should just have no default.

src/cpu/probes/LoopPointManager.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class LoopPointManager(SimObject):
+
+    type = "LoopPointManager"
+    cxx_header = "cpu/probes/looppointmanager.hh"
+    cxx_class = "gem5::LoopPointManager"
+
+    target_count = VectorParam.Int(0, "the target PC count")
+    target_pc = VectorParam.UInt64(0, "the target PC")
```
powerjg: Probably shouldn't have any default values

src/cpu/probes/LoopPointManager.py
```diff
@@ -0,0 +1,16 @@
+from m5.params import *
```
powerjg: Copyright

src/cpu/probes/SConscript
```diff
@@ -0,0 +1,8 @@
+Import('*')
```
powerjg, line 27: copyright

src/cpu/probes/looppoint.cc
```diff
@@ -0,0 +1,42 @@
+#include "cpu/probes/looppoint.hh"
```
powerjg, line 29: copyright

src/cpu/probes/looppoint.cc
```diff
@@ -0,0 +1,42 @@
...
+
+namespace gem5
+{
+
+LoopPoint::LoopPoint(const LoopPointParams &p)
+    : ProbeListenerObject(p),
+    targetPC(p.target_pc),
+    cpuptr(p.core),
+    lpmanager(p.LPmanager)
+{
```
powerjg, line 38: If it's required that certain things are non-null, this is a good place to throw a `fatal` if they are null.

src/cpu/probes/looppoint.cc
```diff
@@ -0,0 +1,42 @@
...
+LoopPoint::check_pc(const Addr& pc)
+{
+    for (int i =0; i<targetPC.size();i++)
+    {
+        if(pc == targetPC[i])
+        {
+            if (lpmanager->check_count(pc))
+                cpuptr->scheduleInstStop(0,1,"simpoint starting point found");
+        }
+    }
```
powerjg:
```suggestion
    for (int i =0; i < targetPC.size(); i++) {
        if (pc == targetPC[i]) {
            if (lpmanager->check_count(pc)) {
                cpuptr->scheduleInstStop(0,1,"simpoint starting point found");
            }
        }
    }
```

src/cpu/probes/looppoint.hh
```diff
@@ -0,0 +1,30 @@
+#ifndef __CPU_PROBES_LOOPPOINT_HH__
```
powerjg, line 29: copyright

src/cpu/probes/looppoint.hh
```diff
@@ -0,0 +1,30 @@
...
+{
+    public:
+        LoopPoint(const LoopPointParams &params);
+        virtual ~LoopPoint();
+        virtual void init();
+        virtual void regProbeListeners();
+        void check_pc(const Addr&);
+
+    private:
+        const std::vector<Addr> targetPC;
```
powerjg: I suggest making this an `unordered_map<Addr,unsigned>`. It should perform better and you won't have to use `check_pc`. Instead, you can set this up at initialization time (maybe through `init()`?)

src/cpu/probes/looppointmanager.cc
```diff
@@ -0,0 +1,57 @@
+#include "cpu/probes/looppointmanager.hh"
```
powerjg, line 29: copyright

src/cpu/probes/looppointmanager.cc
```diff
@@ -0,0 +1,57 @@
...
+            targetCount.insert(std::make_pair(p.target_pc[i], pcCount));
+            counter.insert(std::make_pair(p.target_pc[i],0));
+        } else {
+            std::vector<int>& pcCount = map_itr->second;
+            pcCount.push_back(p.target_count[i]);
+        }
+    }
+    info = simout.create("LoopPointInfo.txt", false);
+    if (!info)
+        fatal("unable to open LoopPoint info txt");
```
powerjg:
```suggestion
    for (int i = 0; i< p.target_pc.size(); i++) {
        auto map_itr = targetCount.find(p.target_pc[i]);
        if (map_itr == targetCount.end()) {
            std::vector<int> pcCount = {p.target_count[i]};
            targetCount.insert(std::make_pair(p.target_pc[i], pcCount));
            counter.insert(std::make_pair(p.target_pc[i],0));
        } else {
            std::vector<int>& pcCount = map_itr->second;
            pcCount.push_back(p.target_count[i]);
        }
    }
    info = simout.create("LoopPointInfo.txt", false);
    if (!info) {
        fatal("unable to open LoopPoint info txt");
    }
```

src/cpu/probes/looppoint.hh
```diff
@@ -0,0 +1,30 @@
...
+#include "cpu/probes/looppointmanager.hh"
+#include "params/LoopPoint.hh"
+#include "sim/probe/probe.hh"
+
+namespace gem5
+{
+
+class LoopPoint : public ProbeListenerObject
+{
+    public:
```
powerjg: Two spaces instead of 4.

src/cpu/probes/looppointmanager.cc
```diff
@@ -0,0 +1,57 @@
+#include "cpu/probes/looppointmanager.hh"
+
+namespace gem5
+{
+LoopPointManager::LoopPointManager(const LoopPointManagerParams &p)
+    : SimObject(p)
+{
+    for (int i = 0; i< p.target_pc.size(); i++)
```
powerjg: Can you add some comments here explaining this code?

src/cpu/probes/looppointmanager.cc
```diff
@@ -0,0 +1,57 @@
...
+    std::vector<int>& targetcount = target_itr->second;
+    for (std::vector<int>::iterator iter = targetcount.begin(); iter < targetcount.end(); iter++)
+    {
+        if(*iter==count)
+        {
+            targetcount.erase(iter);
+            *info->stream() << curTick() << " : " << pc << " : " << count << " \n "; 
+            return true;
+        }
+    }
```
powerjg:
```suggestion
    for (std::vector<int>::iterator iter = targetcount.begin(); iter < targetcount.end(); iter++) {
        if(*iter==count) {
            targetcount.erase(iter);
            *info->stream() << curTick() << " : " << pc << " : " << count << " \n "; 
            return true;
        }
    }
```

src/cpu/probes/looppointmanager.cc
```diff
@@ -0,0 +1,57 @@
...
+    int& count = count_itr-> second;
+    count += 1;
+    auto target_itr = targetCount.find(pc);
+    std::vector<int>& targetcount = target_itr->second;
+    for (std::vector<int>::iterator iter = targetcount.begin(); iter < targetcount.end(); iter++)
+    {
+        if(*iter==count)
+        {
+            targetcount.erase(iter);
+            *info->stream() << curTick() << " : " << pc << " : " << count << " \n ";
```
powerjg: I would exit here. I think you can use a function like `exitSimLoop`. See `exit_event.hh` or something like that.

src/cpu/probes/looppoint.hh
powerjg: Actually, I think this should be an `unordered_set`

src/cpu/probes/looppointmanager.hh
```diff
@@ -0,0 +1,32 @@
+#ifndef __CPU_PROBES_LOOPPOINTMANAGER_HH__
```
powerjg, line 29: copyright

src/cpu/probes/looppointmanager.hh
```diff
@@ -0,0 +1,32 @@
...
+        virtual ~LoopPointManager();
+        virtual void init();
+        bool check_count(uint64_t);
+
+    private:
+        // const std::vector<int> targetCount;
+        // std::vector<int> counter;
+        OutputStream *info;
+        std::unordered_map<uint64_t, std::vector<int>> targetCount;
+        std::unordered_map<uint64_t,int> counter;
```
powerjg: Add some comments explaining these.

src/cpu/probes/looppointmanager.hh
```diff
@@ -0,0 +1,32 @@
...
+namespace gem5
+{
+
+class LoopPointManager : public SimObject
+{
+    public:
+        LoopPointManager(const LoopPointManagerParams &params);
+        virtual ~LoopPointManager();
+        virtual void init();
+        bool check_count(uint64_t);
```
powerjg: Docstring and also give the variable a name :)

src/cpu/probes/LoopPoint.py
studyztp: I got python error when building gem5 if I don't put NULL here. I think NULL is valid when Param is a SimObject.

src/cpu/probes/LoopPoint.py
studyztp: I changed the default value from ```0``` to ```[]```. ```[]``` should be the valid default value for VectorParam. If I don't set a default value, it raised python errors when I tried to build gem5.

src/cpu/probes/LoopPoint.py
powerjg: Odd! I would not expect that error. @BobbyRBruce, we should document this behavior or we should figure out if it's a bug. It surprises me that a default value is *required* for SimObject params 

What I meant by valid is whether or not it would be valid for the user to specify NULL. I think not. I believe there would be a fatal or segfault raised. That said, if gem5 requires a default, then that's how it is.

studyztp:
@powerjg Thank you for the feedbacks! I fixed the issues and added the comments.

src/cpu/probes/LoopPoint.py
powerjg: Like the comment above, this should not be the case. Maybe it is, but I think it's a bug in gem5.

I can find plenty of examples both where there is no default for the `Param.<SimObject>` and when there's no default for a `VectorParam`. 

I'm not saying I don't believe you, but it's very strange! :)

src/cpu/probes/LoopPoint.py
studyztp: It could be I was doing something wrong with other parts. I'll test with only the Param SimObject without default value now.

src/cpu/probes/LoopPoint.py
```diff
@@ -0,0 +1,39 @@
...
+from m5.objects.Probe import ProbeListenerObject
+
+
+class LoopPoint(ProbeListenerObject):
+
+    type = "LoopPoint"
+    cxx_header = "cpu/probes/looppoint.hh"
+    cxx_class = "gem5::LoopPoint"
+
+    target_pc = VectorParam.UInt64("the target PC")
```
powerjg:
```suggestion
    target_pc = VectorParam.Addr("the target PC")
```

Is `Addr` a valid param type? I think it is.

src/cpu/probes/LoopPointManager.py
```diff
@@ -0,0 +1,49 @@
...
+from m5.objects import SimObject
+
+class LoopPointManager(SimObject):
+
+    type = "LoopPointManager"
+    cxx_header = "cpu/probes/looppointmanager.hh"
+    cxx_class = "gem5::LoopPointManager"
+
+    target_count = VectorParam.Int("the target PC count")
+    target_pc = VectorParam.UInt64("the target PC")
```
powerjg:
```suggestion
    target_pc = VectorParam.Addr("the target PC")
```

src/cpu/probes/LoopPointManager.py
```diff
@@ -0,0 +1,49 @@
...
+    region_id = VectorParam.Int("the simulation region id")
+    relative_pc = VectorParam.UInt64("the relative PC of the target PC")
```
powerjg:
```suggestion
    relative_pc = VectorParam.Addr("the relative PC of the target PC")
```

src/cpu/probes/LoopPointManager.py
```diff
@@ -0,0 +1,49 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.params import *
+from m5.objects import SimObject
+
+class LoopPointManager(SimObject):
```
powerjg: Small nit: I would put this class in the `LoopPoint.py` file.

src/python/gem5/utils/looppoint.py
```diff
@@ -0,0 +1,219 @@
...
+from gem5.components.processors.abstract_core import AbstractCore
+from m5.objects.LoopPointManager import LoopPointManager
+import csv
+import re
+from abc import ABCMeta
+        
+        
+class BaseLoopPoints:
+    
+    __metaclass__ = ABCMeta
```
powerjg: I don't think this is needed.

src/python/gem5/components/processors/base_cpu_core.py
```diff
@@ -165,3 +170,15 @@ def set_inst_stop_any_thread(self, inst: int, init: bool) -> None:
             self.core.max_insts_any_thread = inst
         else:
             self.core.scheduleInstStopAnyThread(inst)
+            
+    @overrides(AbstractCore)
+    def addLoopPointProbe(
+        self, 
+        targetpc: List[AbstractCore],
```
powerjg:
```suggestion
        targetpc: List[int],
```

src/python/gem5/components/processors/base_cpu_core.py
```diff
@@ -25,6 +25,8 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from typing import Optional, List
+
+from m5.params import UInt64
```
powerjg: Remove

src/cpu/probes/LoopPointManager.py
```diff
@@ -0,0 +1,49 @@
...
+    cxx_header = "cpu/probes/looppointmanager.hh"
+    cxx_class = "gem5::LoopPointManager"
+
+    target_count = VectorParam.Int("the target PC count")
+    target_pc = VectorParam.UInt64("the target PC")
+    region_id = VectorParam.Int("the simulation region id")
+    relative_pc = VectorParam.UInt64("the relative PC of the target PC")
+    relative_count = VectorParam.Int("the relative PC count of the target PC")
+
+    def setup(self, cpulist):
```
powerjg: Is this function required? I would rather not have general python in the "SimObject declaration files" if it's avoidable.

src/python/gem5/utils/looppoint.py
```diff
@@ -0,0 +1,219 @@
...
+    def setup_cpu(
+        self, 
+        cpuList: List[AbstractCore], 
+        ) -> None:
+        self._looppointManager.target_count = self._checkpointCount
+        self._looppointManager.target_pc = self._checkpointPC
+        self._looppointManager.region_id = self._regionID
+        self._looppointManager.relative_pc = self._relativePC
+        self._looppointManager.relative_count = self._relativeCount
+        self._looppointManager.setup(cpuList)
```
powerjg:
```suggestion
    def setup_processor(
        self, 
        processor: List[AbstractProcessor], 
        ) -> None:
        self._looppointManager.target_count = self._checkpointCount
        self._looppointManager.target_pc = self._checkpointPC
        self._looppointManager.region_id = self._regionID
        self._looppointManager.relative_pc = self._relativePC
        self._looppointManager.relative_count = self._relativeCount
        for core in processor.get_cores():
            core.addLoopPointProbe(self._checkpointPC + self._relativePC, self._looppointManager)
```

Note, I'm not very confident on the last line.

Whether you go with this suggestion or something closer to the original... I suggest using the name "core" or "processor" and avoiding the name "cpu" for variables.

src/cpu/probes/looppoint.cc
```diff
@@ -0,0 +1,75 @@
...
+}
+
+LoopPoint::~LoopPoint()
+{}
+
+void
+LoopPoint::init()
+{}
```
powerjg: You can remove these empty functions

src/cpu/probes/looppoint.hh
```diff
@@ -0,0 +1,76 @@
...
+
+namespace gem5
+{
+
+class LoopPoint : public ProbeListenerObject
+{
+  public:
+    LoopPoint(const LoopPointParams &params);
+    virtual ~LoopPoint();
+    virtual void init();
```
powerjg: Not needed :)

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -111,3 +112,37 @@ def set_se_binary_workload(
 
         # Set whether to exit on work items for the se_workload
         self.exit_on_work_items = exit_on_work_items
+
+    def set_se_looppoint_workload(
```
powerjg, line 116: I think we changed the way we're doing simpoints. Can you follow what's done here; https://gem5-review.googlesource.com/c/public/gem5/+/64432

powerjg (commented):
A few comments. I think you might have closed this, but the comments still stand :).

Good work so far!

studyztp:
I'll fix them and move them to Gerrit. Thanks for the great comments! :)

PR #257, opened 2023-02-13
Change for split requests

aakahlow:
Has not been extensively tested yet.

### DIFF
--- a/src/cpu/testers/dr_trace_player/trace_player.cc
+++ b/src/cpu/testers/dr_trace_player/trace_player.cc
@@ -165,9 +165,7 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
 bool
 DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
 {
-    PacketPtr pkt, split_pkt;
-
-    std::tie(pkt, split_pkt) = getPacket(mem_ref);
+    auto [pkt, split_pkt] = getPacket(mem_ref);
 
     if (!retrySplitPkt) {
         // we should not be sending the first pkt again if
@@ -180,18 +178,20 @@ DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
             if (stats.memStallStart == 0) {
                 stats.memStallStart = curTick();
             }
+            delete pkt;
+
+            // return true if we have to stall on the first pkt
+            // irrespective of if this is a split req
+            return true;
+        } else {
             numOutstandingMemReqs++;
             stats.outstandingMemReqs.sample(numOutstandingMemReqs);
-
-            delete pkt;
-            if (split_pkt == nullptr) {
-            // if this is not a split pkt req,
-            // we can return back here
-            return true;
-            }
-        } else {
             stats.latencyTracker[pkt] = curTick();
-            return false;
+            if (split_pkt == nullptr) {
+                // if this is not a split req, we can
+                // return here
+                return false;
+            }
         }
     }
 
@@ -202,22 +202,23 @@ DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
         if (stats.memStallStart == 0) {
             stats.memStallStart = curTick();
         }
-        numOutstandingMemReqs++;
-        stats.outstandingMemReqs.sample(numOutstandingMemReqs);
-
         delete split_pkt;
         // also remember that we only need to retry on second part of
         // the split pkt
         retrySplitPkt = true;
         return true;
     } else {
+        numOutstandingMemReqs++;
+        stats.outstandingMemReqs.sample(numOutstandingMemReqs);
         stats.latencyTracker[split_pkt] = curTick();
         retrySplitPkt = false;
+        // At this point, we are sure that both pkts of the split req
+        // are received by the port
         return false;
     }
 }
 
-std::pair<PacketPtr, PacketPtr>
+std::tuple<PacketPtr, PacketPtr>
 DRTracePlayer::getPacket(DRTraceReader::TraceRef &mem_ref)
 {
     Request::Flags flags = Request::PHYSICAL;
@@ -292,7 +293,7 @@ DRTracePlayer::getPacket(DRTraceReader::TraceRef &mem_ref)
         }
     }
 
-    return std::make_pair(pkt, split_pkt);
+    return {pkt, split_pkt};
 }
 
 Port &
--- a/src/cpu/testers/dr_trace_player/trace_player.cc
+++ b/src/cpu/testers/dr_trace_player/trace_player.cc
@@ -179,6 +179,7 @@ DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
                 stats.memStallStart = curTick();
             }
             delete pkt;
+            delete split_pkt;
 
             // return true if we have to stall on the first pkt
             // irrespective of if this is a split req
@@ -193,6 +194,11 @@ DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
                 return false;
             }
         }
+    } else {
+        // we should delete the first pkt here
+        // if we are only trying to resend the
+        // second pkt
+        delete pkt;
     }
 
     DPRINTF(DRTrace, "Trying to send split %s\n", split_pkt->print());
--- a/src/cpu/testers/dr_trace_player/trace_player.cc
+++ b/src/cpu/testers/dr_trace_player/trace_player.cc
@@ -167,24 +167,53 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
 bool
 DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
 {
-    PacketPtr pkt = getPacket(mem_ref);
+    PacketPtr pkt, split_pkt;
 
-    DPRINTF(DRTrace, "Trying to send %s\n", pkt->print());
+    std::tie(pkt, split_pkt) = getPacket(mem_ref);
 
-    if (!port.sendTimingReq(pkt)) {
-        DPRINTF(DRTrace, "Failed to send pkt\n");
+    if (!retrySplitPkt) {
+        // we should not be sending the first pkt again if
+        // we are retrying only on the second part of the
+        // previously stalled request
+        DPRINTF(DRTrace, "Trying to send %s\n", pkt->print());
+
+        if (!port.sendTimingReq(pkt)) {
+            DPRINTF(DRTrace, "Failed to send pkt\n");
+            if (stats.memStallStart == 0) {
+                stats.memStallStart = curTick();
+            }
+            delete pkt;
+            if (!split_pkt) {
+            // if this is not a split pkt req,
+            // we can return back here
+            return true;
+            }
+        } else {
+            stats.latencyTracker[pkt] = curTick();
+            return false;
+        }
+    }
+
+    DPRINTF(DRTrace, "Trying to send split %s\n", split_pkt->print());
+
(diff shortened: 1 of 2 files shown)


src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +167,53 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
-        delete pkt;
+        retrySplitPkt = false;
+        delete split_pkt;
         return true;
     } else {
-        stats.latencyTracker[pkt] = curTick();
+        stats.latencyTracker[split_pkt] = curTick();
+        // also remember that we only need to retry on second part of
+        // the split pkt
+        retrySplitPkt = true;
```
powerjg, line 221: I think this is backwards. The `true` should go in the if statement above when the `sendTimingReq` fails

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -230,9 +261,35 @@ DRTracePlayer::getPacket(DRTraceReader::TraceRef &mem_ref)
         }
     }
 
-    return pkt;
-}
+    PacketPtr split_pkt = NULL;
```
powerjg:
```suggestion
    PacketPtr split_pkt = nullptr;
```

powerjg (changes requested):
One thing that's missing is updates to `numOutstandingMemReqs`. Right now, this is going to go below 0 (we should probably add an assert on (new) line 320.

Right now, there is just one memory request added for a split packet, but we are subtracting twice. We should either only subtract when we receive both packets or we should add twice. I think the latter would be better.

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,59 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
 bool
 DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
 {
-    PacketPtr pkt = getPacket(mem_ref);
+    PacketPtr pkt, split_pkt;
+
+    std::tie(pkt, split_pkt) = getPacket(mem_ref);
```
powerjg:
```suggestion
   auto [pkt, split_pkt] = getPacket(mem_ref);
```

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -230,9 +265,35 @@ DRTracePlayer::getPacket(DRTraceReader::TraceRef &mem_ref)
...
+            split_pkt->dataDynamic(split_pkt_data);
+
+            if (cmd.isWrite()) {
+                std::fill_n(split_pkt_data, split_req->getSize(),
+                                            (uint8_t)requestorId);
+            }
+        }
+    }
+
+    return std::make_pair(pkt, split_pkt);
```
powerjg: I think the following works in modern C++.

```suggestion
    return {pkt, split_pkt};
```

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,59 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
     } else {
-        stats.latencyTracker[pkt] = curTick();
+        stats.latencyTracker[split_pkt] = curTick();
+        retrySplitPkt = false;
         return false;
     }
 }
 
-PacketPtr
+std::pair<PacketPtr, PacketPtr>
```
powerjg: I *think* this is right, but you should double check me :).
```suggestion
std::tuple<PacketPtr, PacketPtr>
```

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,59 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
+        // we are retrying only on the second part of the
+        // previously stalled request
+        DPRINTF(DRTrace, "Trying to send %s\n", pkt->print());
+
+        if (!port.sendTimingReq(pkt)) {
+            DPRINTF(DRTrace, "Failed to send pkt\n");
+            if (stats.memStallStart == 0) {
+                stats.memStallStart = curTick();
+            }
+            numOutstandingMemReqs++;
```
powerjg: Why do we increment the outstanding requests when the packet is failed?

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,59 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
+            stats.outstandingMemReqs.sample(numOutstandingMemReqs);
+
+            delete pkt;
+            if (split_pkt == nullptr) {
+            // if this is not a split pkt req,
+            // we can return back here
+            return true;
```
powerjg, line 190: I think we should delete the other packet and return when the first packet fails.

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,59 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
+            }
+        } else {
+            stats.latencyTracker[pkt] = curTick();
+            return false;
+        }
+    }
 
-    DPRINTF(DRTrace, "Trying to send %s\n", pkt->print());
+    DPRINTF(DRTrace, "Trying to send split %s\n", split_pkt->print());
```
powerjg, line 208: We should only send the second packet when the first packet successfully sends.

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,59 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
 
-    if (!port.sendTimingReq(pkt)) {
-        DPRINTF(DRTrace, "Failed to send pkt\n");
+    if (!port.sendTimingReq(split_pkt)) {
+        DPRINTF(DRTrace, "Failed to send pkt (split pkt) \n");
         if (stats.memStallStart == 0) {
             stats.memStallStart = curTick();
         }
-        delete pkt;
+        numOutstandingMemReqs++;
```
powerjg: Again, why increcement when it fails?

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,59 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
+            if (split_pkt == nullptr) {
+            // if this is not a split pkt req,
+            // we can return back here
+            return true;
+            }
+        } else {
+            stats.latencyTracker[pkt] = curTick();
+            return false;
+        }
+    }
```
powerjg, line 206: What happens if `retrySplitPacket` is true?

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: done

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: done

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: done

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: done

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: sorry, I pushed the wrong version earlier

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: if retrySplitPacket is true, that means we are retrying only the second pkt and therefore, should not do anything for the first pkt.

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,60 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
 bool
 DRTracePlayer::trySendMemRef(DRTraceReader::TraceRef &mem_ref)
 {
-    PacketPtr pkt = getPacket(mem_ref);
+    auto [pkt, split_pkt] = getPacket(mem_ref);
+
+    if (!retrySplitPkt) {
+        // we should not be sending the first pkt again if
+        // we are retrying only on the second part of the
+        // previously stalled request
```
powerjg: This comment is for the other side of the branch, not this side.

src/cpu/testers/dr_trace_player/trace_player.cc
powerjg: Let me know if you disagree with this comment.

src/cpu/testers/dr_trace_player/trace_player.cc
powerjg: Don't we need to delete the first packet?

src/cpu/testers/dr_trace_player/trace_player.cc
powerjg: This has not been addressed.

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: Yes, that's right. Pushed  a fix

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: Pushed a change to delete the split_pkt as well when the first pkt fails

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: If I have understood correctly, the place of the comment should be changed?

src/cpu/testers/dr_trace_player/trace_player.cc
aakahlow: I think I have not fully understood this concern/comment. This DPRINTF statement will be executed only if we successfully send out the first packet.

powerjg: Sorry for the confusion. This wasn't a comment on the DPRINTF. It's been addressed.

src/cpu/testers/dr_trace_player/trace_player.cc
powerjg: Correct.

aakahlow: Added/fixed comments to hopefully make things better and not worse

src/cpu/testers/dr_trace_player/trace_player.cc
```diff
@@ -167,24 +165,73 @@ DRTracePlayer::executeMemInst(DRTraceReader::TraceRef &mem_ref)
...
-    PacketPtr pkt = getPacket(mem_ref);
+    // split_pkt will be nullptr if not a split req
+    auto [pkt, split_pkt] = getPacket(mem_ref);
+
+    // ensure that currently we are not in process of
+    // retrying to send the second part of a previously
+    // stalled request to avoid duplicate first pkt in
+    // the memory system.
+    // Also, the assumption is that we cannot start a
+    // new memory request in parallel.
```
powerjg, line 176: I'm not sure how to interpret this statement. Otherwise, it looks good.

PR #267, opened 2023-10-03
William

Will-Shaddix:
For Jason to review, MessageQueue changes are in message_queue.cc,  .hh, Msg_Map_process_test.py is my test config,  mapped_queue.cpp is what was used to create the binary I am running in the config


src/mem/DRAMInterface.py
```diff
@@ -1806,3 +1806,101 @@ class LPDDR5_6400_1x16_8B_BL32(LPDDR5_6400_1x16_BG_BL32):
     tCCD_L = "0ns"
     tRRD_L = "0ns"
     tWTR_L = "0ns"
+
+
+class LLM(DRAMInterface):
```
powerjg: Can you drop LLM changes from this PR?

src/mem/MessageQueue.py
```diff
@@ -0,0 +1,17 @@
+from m5.params import *
+from m5.proxy import * 
+from m5.objects.ClockedObject import ClockedObject
+
+
+class MessageQueue(ClockedObject):
+    type = 'MessageQueue'
+    cxx_header = "mem/message_queue.hh"
+    cxx_class = "gem5::MessageQueue"
+    # can either have a vector of ports connecting to the cores, or have a vector of write ports and 1 read port
```
powerjg, line 10: make sure you install `pre-commit` and follow the gem5 style guide (in this case lines should be < 80 chars)

src/mem/message_queue.hh
```diff
@@ -0,0 +1,137 @@
...
+                virtual bool recvTimingReq(PacketPtr pkt);
+                virtual Tick recvAtomic(PacketPtr pkt);
+                virtual void recvFunctional(PacketPtr pkt);
+                virtual void recvRespRetry();
+
+
+            
+        };
+
+        std::deque<std::tuple<uint32_t, uint32_t, uint32_t, Tick>> queue; //Could be address or vertexID, vertexID might be easier
```
powerjg: This seems like a very complex datastructure. Maybe it should be a map? A `deque` of a complex `tuple` doesn't seem like the right solution. In other words, this is a bad code smell :)

src/mem/message_queue.hh
```diff
@@ -0,0 +1,137 @@
...
+        
+        uint32_t queueSize;
+        AddrRange my_range;
```
powerjg: Please follow the gem5 coding style. In this case, you need to use `camelCase`
```suggestion
        AddrRange myRange;
```

src/mem/message_queue.hh
```diff
@@ -0,0 +1,137 @@
...
+        // std::vector<RespPort> corePorts; 
+        RespPort corePorts;//= new RespPort();
```
powerjg, line 123: This can't be `new` since the `RespPort` is not a pointer.

src/mem/message_queue.hh
```diff
@@ -0,0 +1,137 @@
...
+#define __MESSAGE_QUEUE_HH__
+
+#include "params/MessageQueue.hh"
+#include "sim/clocked_object.hh"
+#include "mem/port.hh"
+#include "base/addr_range.hh"
+#include "sim/system.hh"
+#include "mem/packet_access.hh"
+#include "debug/MessageQueue.hh"
+#include "dev/io_device.hh"
```
powerjg, line 11: Is this used?

src/mem/message_queue.cc
```diff
@@ -0,0 +1,333 @@
+#include "mem/message_queue.hh"
+#include <cstdio>
+
+// /src/mem/message_queue.hh
+
+namespace gem5{
+
+
+
+    MessageQueue::MessageQueue(const MessageQueueParams &p) :
```
powerjg, line 11: Should not have leading spaces. Please follow the style guide.

configs/William/mapped_queue.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+
+
+int main(int argc, char** argv){
+
+    char data[64];
+    int num_bytes = 64;
+    char* to_write = (char*)"as"; // needs to be at least 1 byte of data in before calling mmap
+    
+    char* mapped_area; //could replace char with data type/tuple?
```
powerjg: This shouldn't be a `char*``since this is not characters.

It could be a `uint8_t*` if you want to just have "data."

configs/William/mapped_queue.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    int buffer_size = 4096;
+    int buffer_addr = open("buffer_addr", O_RDWR | O_CREAT, S_IRWXU | S_IRWXO);
+    int y = write(buffer_addr, to_write, 2);
```
powerjg: What is this?

configs/William/mapped_queue.cpp
```diff
@@ -0,0 +1,58 @@
...
+    // mapped_area = (char*)mmap((void *)0x1000000, buffer_size, PROT_READ | PROT_WRITE, MAP_SHARED, buffer_addr, 0);  // Use this line to run locally
+    mapped_area = (char*)mmap((void *)0x1000000, buffer_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_FIXED , buffer_addr, 0);  // MAP_FIXED only for use in gem5!
```
powerjg: `0x1000000` should be a constant.

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+
+This config file assumes that the x86 ISA was built.
+"""
+
+# import the m5 (gem5) library created when gem5 is built
+import m5
+# import all of the SimObjects
+from m5.objects import *
+from m5.util.convert import *
+from caches import *
```
powerjg: I think that you forgot to include this file.

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+
+# import the SimpleOpts module
+from common import SimpleOpts
```
powerjg: Don't use `SimpleOpts` (and remove since unused). It was a mistake for me to push that module :)

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+import math
+
+import argparse
+
+
+parser = argparse.ArgumentParser()
```
powerjg: Why do you have a parser without any options?

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+options.duration = int(toLatency("100us") * 1e12)
+options.min_addr = 0
+#options.min_addr = toMemorySize(str(1 * options.num_chnls) + 'MB')
+options.max_addr = toMemorySize(str(4 * options.num_chnls) + 'KiB')
+options.max_addr = toMemorySize(str(1024 * options.num_chnls) + 'MB')
+options.injection_rate = 1
+injection_period = int((1e12 * options.block_size) /
+                    (options.injection_rate * 1073741824))
+options.min_period = injection_period
+options.max_period = injection_period
```
powerjg: This is very strange... This is not how you're supposed to use `argparse`. See https://docs.python.org/3/howto/argparse.html

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+system.mem_ranges = [AddrRange(0, '255MB'), AddrRange('255MB', '256MB')] # Create an address range
+# system.mem_ranges = [AddrRange(0, '1023MB'), AddrRange('1023MB', '1024MB')] # Create an address range
+
+# system.mem_ranges = [AddrRange(0, '1MB'), AddrRange('1MB', '1024MB')] # Create an address range
+
+addr_range = system.mem_ranges[0]
+queue_ranges = [system.mem_ranges[1]]
+
+# system.membus = SystemXBar(width = 64, max_routing_table_size = 16777216)
+system.membus = IOXBar(width = 64)
```
powerjg: The membus (i.e., the point of coherence) can't be an `IOXBar`, it must be a `SystemXBar`

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+system = System()
+
+# Set the clock fequency of the system (and all of its children)
+system.clk_domain = SrcClockDomain()
+system.clk_domain.clock = '1GHz'
+system.clk_domain.voltage_domain = VoltageDomain()
+
+# Set up the system
+system.mem_mode = 'timing'               # Use timing accesses   AddrRange('1023MB', '1047556KiB')]
+system.mem_ranges = [AddrRange(0, '255MB'), AddrRange('255MB', '256MB')] # Create an address range
```
powerjg, line 52: The *system's* address range should just be 0-256MiB (or whatever the **SYSTEM'S** address range is).

Then, you should set a different address range for the message queue.

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+# system.cpu.icache = L1ICache()
+# system.cpu.icache_port = system.cpu.icache.cpu_side
+# system.cpu.icache.mem_side = system.membus.cpu_side_ports
+
+#queue_ranges = [AddrRange('0', '4KiB')] # address range for queue
+# queue_ranges = [system.mem_ranges[1]]
+
+# system.bridge = Bridge(ranges=queue_ranges)
+
+system.msg_queue = MessageQueue(my_range=queue_ranges[0], queueSize=18)
```
powerjg, line 77: Shouldn't you have a message queue per core?

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+                        xorHighBit = 0,
+                        intlvBits = intlv_bits,
+                        intlvMatch = chnl)
+            ctrl = MemCtrl()
+            ctrl.dram = interface
+
+            #ctrl.dram.null = True
+            #ctrl.dram.addr_mapping = addr_map
+            #ctrl.dram.page_policy = page_policy
+            mem_ctrls.append(ctrl)
```
powerjg, line 105: You should simplify this for now. You should simply have one memory channel.

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+# below here also for cpu
+# Create a process for a simple "Hello World" application
+process = Process()
+# process.map(vaddr=0x1000000, paddr=0x3CF9BDC0, size=4096, cacheable=False)
+# Set the command
+# grab the specific path to the binary
+thispath = os.path.dirname(os.path.realpath(__file__))
+
+binpath = os.path.join(thispath, '../../',
+                       'tests/test-progs/hello/bin/x86/linux/mapped_queue_fixed_private')
```
powerjg, line 133: This doesn't have to be in test-progs, and you should avoid checking in binaries.

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+# binpath = os.path.join(thispath, '../../',
+#                        'tests/test-progs/hello/bin/x86/linux/no_tuple')
+# cmd is a list which begins with the executable (like argv)
+process.cmd = [binpath]
+# Set the cpu to use the process as its workload and create thread contexts
+system.cpu.workload = process
+# system.cpu.workload.map(vaddr=0x1000000, paddr=0x3CF9BDC0, size=4096)
```
powerjg, line 140: This `paddr` (`0x3CF9BDC0`) is what you need to set your `addr_range` for the message queue.

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+system.workload = SEWorkload.init_compatible(binpath)
+
+# set up the root SimObject and start the simulation
+root = Root(full_system = False, system = system)
+# instantiate all of the objects we've created above
+m5.instantiate()
+
+# process.map(vaddr=0x10000, paddr=0x3CF9BDC0, size=4096, cacheable=False)
+
+process.map(vaddr=0x1000000, paddr=0xFF00000, size=4096, cacheable=False)
```
powerjg: Or maybe the above comment should be here?

configs/William/Msg_Map_process_test.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+print("Beginning simulation!")
+exit_event = m5.simulate()
+print('Exiting @ tick %i because %s' % (m5.curTick(), exit_event.getCause()))
```
powerjg, line 157: Don't use `%` for formatting. Use [f-strings](https://peps.python.org/pep-0498/)

Another link: <https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/#>

powerjg (commented):
Ok, thanks for putting this together. I have lots of small comments. Please take the time to learn gem5's style guide and take the time to write code that is up to our standards. Use stack overflow **and github copilot!!** to find the "right way" to write code.

For the actual problem that you're facing... the code is missing some pieces. I can try to find some time to sit down with you and work on this. Maybe tomorrow afternoon?

configs/William/Msg_Map_process_test.py
Will-Shaddix: I set the Addr Range of Message Queue to be 255MB - 256MB which I believe 255 MB is 0xFF00000, unless I calculated that wrong.  I know the size doesn't really match up to 1 MB but I figured the pointer should be pointing to 255 MB.

configs/William/mapped_queue.cpp
Will-Shaddix: I noticed when trying to get mmap to work, the file needed to have something in it before i called mmap, otherwise  it would throw an error because its trying to mmap 0 bytes, so i was just testing it by giving it some random characters to start with.

PR #275, opened 2024-04-09
repo: created a new branch for review

kaustav-goswami:
A setup to do disaggregated memory simulation using gem5 and SST.


disaggregated_memory/boards/riscv_main_board.py
```diff
@@ -0,0 +1,596 @@
...
+
+        super().__init__(
+            clk_freq=clk_freq,
+            processor=processor,
+            memory=local_memory,
+            cache_hierarchy=cache_hierarchy,
+        )
+
+        self.local_memory = local_memory
+        self.remote_memory = remote_memory
```
powerjg, line 148: Why renaming these here? Above you called them `_local_memory` and here it's renamed. Why?

disaggregated_memory/boards/riscv_main_board.py
```diff
@@ -0,0 +1,596 @@
...
+        # memory can start and end anywhere as long as it is consistent
+        # with the dtb.
+        self._local_mem_ranges = [
+            AddrRange(start=0x80000000, size=local_mem_size)
+        ]
+
+        # The remote memory starts anywhere after the local memory ends. We
+        # rely on the user to start and end this range.
+        self._remote_mem_ranges = [
+            self.get_remote_memory().get_mem_ports()[0][0]
```
powerjg, line 252: Why `[0][0]`? Should the interface to `get_mem_ports` be changed?

disaggregated_memory/boards/arm_main_board.py
```diff
@@ -0,0 +1,469 @@
...
+    single board to make the boards compatible with both gem5 and SST.
+
+    **Limitations**
+    * kvm is only supported in a gem5-only setup.
+
+    @params
+    TODO
+    """
+
+    __metaclass__ = ABCMeta
```
powerjg: Why is this here?

disaggregated_memory/memories/external_remote_memory_v2.py
```diff
@@ -0,0 +1,215 @@
...
+    fatal,
+    warn,
+)
+
+from gem5.components.boards.abstract_board import AbstractBoard
+from gem5.components.memory.memory import AbstractMemorySystem
+from gem5.utils.override import overrides
+
+
+class ExternalRemoteMemoryV2(AbstractMemorySystem):
```
powerjg: Don't call this V2

disaggregated_memory/memories/external_remote_memory_v2.py
```diff
@@ -0,0 +1,215 @@
...
+        """This class has to be initialized using either size or memory ranges.
+
+        Args:
+            size (str, optional): Size. Defaults to None.
+            addr_range (AddrRange, optional): Address Range. Defaults to None.
+            link_latency (Tick, optional): Additional latency. Defaults to None
+        """
+        super().__init__()
+
+        # TODO: marked for deletion
```
powerjg: Then delete it :)

disaggregated_memory/memories/external_remote_memory_v2.py
```diff
@@ -0,0 +1,215 @@
...
+                    self.outgoing_request_bridge.physical_address_ranges[
+                        0
+                    ].size()
+                )
+                self._set_using_addr_ranges = True
+            # The size will be setup in the board in case ranges are not given
+            # by the user.
+            else:
+                # There is no range information provided by the user. Depending
+                # upon the ISA, we have to fix the address.
```
powerjg: If it's ISA/system dependent, then do it in the Board class, not the memory class!

disaggregated_memory/memories/external_remote_memory_v2.py
```diff
@@ -0,0 +1,215 @@
...
+        #     self._link_latency = link_latency
+
+    def get_size(self):
+        return self._size
+
+    def is_xbar_required(self):
+        # If an XBar is required, it should be added in the connect_things to
+        # avoid initializing an orphan node.
+        raise NotImplementedError("This is a deprecated method for v2")
+        return self._xbar_required
```
powerjg: delete

disaggregated_memory/memories/external_remote_memory_v2.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def get_link_latency(self):
+        raise NotImplementedError("This is a deprecated method for v2")
+        assert self._xbar_required == True
+        return self._link_latency
```
powerjg: delete

disaggregated_memory/configs/riscv-main.py
```diff
@@ -0,0 +1,293 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+This script shows an example of running a full system ARM Ubuntu boot
+simulation using the gem5 library. This simulation boots Ubuntu 20.04 using
+1 TIMING CPU cores and executes `STREAM`. The simulation ends when the
+startup is completed successfully.
+
+* This script has to be executed from SST
```
powerjg: Is this still true?

disaggregated_memory/configs/riscv-main.py
```diff
@@ -0,0 +1,293 @@
...
+    processor=processor,
+    local_memory=local_memory,
+    remote_memory=remote_memory,
+    cache_hierarchy=cache_hierarchy,
+)
+
+# commands to execute to run the simulation.
+mount_cmd = ["mount -t sysfs - /sys;", "mount -t proc - /proc;"]
+
+warn("The command list to execute has to be manually set!")
```
powerjg, line 186: BTW, this is the whole point of the "workload" objects!

disaggregated_memory/configs/riscv-main.py
```diff
@@ -0,0 +1,293 @@
...
+# Since we are using kvm to boot the system, we can boot the system with
+# systemd enabled!
+cmd = (
+    mount_cmd
+    + ["m5 --addr=0x10010000 exit;"]
+    + local_stream
+    + interleave_stream
+    + remote_stream
+    + ["m5 --addr=0x10010000 exit;"]
+)
```
powerjg: The comment doesn't make sense with the code.

disaggregated_memory/configs/riscv-main.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+# define on_exit_event
+def handle_exit():
+    yield True  # Stop the simulation. We're done.
+
+
+# Here are the different scenarios:
+# no checkpoint, run everything in gem5
+if args.take_ckpt == "True":
```
powerjg, line 268: We can find a better way to communicate "is this executing in SST".

disaggregated_memory/configs/arm-main.py
```diff
@@ -0,0 +1,299 @@
+# Copyright (c) 2023-24 The Regents of the University of California
```
powerjg, line 1: WHy is this file separate from riscv-main.py? The only difference is the board and the workload. I think you could factor out all of the common code!

ext/sst/sst_responder_subcomponent.cc
```diff
@@ -79,14 +80,22 @@ SSTResponderSubComponent::setOutputStream(SST::Output* output_)
 {
     output = output_;
 }
-
+/*
```
powerjg: Delete unused code

ext/sst/sst_responder_subcomponent.cc
```diff
@@ -99,17 +108,68 @@ SSTResponderSubComponent::handleTimingReq(
...
+    if (phase == 0) {
+        // Added support for MPI send and recv. We have to split and send
+        // gem5's data in phases to SST.
+        // get the size of this memory.
+        // We are using a MemBackdoor to get the data to restore from gem5.
+        gem5::MemBackdoorPtr data;
+        responseReceiver->getBackdoor(data);
+        assert(data->readable());
+
+        // this can be a huge number
```
powerjg: What does this comment mean???

ext/sst/sst_responder_subcomponent.cc
```diff
@@ -238,11 +309,59 @@ SSTResponderSubComponent::handleRecvRespRetry()
         responseQueue.pop();
 }
 
+// void
```
powerjg, line 305: Delete unused code

ext/sst/sst_responder_subcomponent.cc
```diff
@@ -238,11 +309,59 @@ SSTResponderSubComponent::handleRecvRespRetry()
...
+    switch((gem5::MemCmd::Command)pkt->cmd.toInt()) {
+        case gem5::MemCmd::WriteReq: {
+            std::vector<uint8_t> data(ptr, ptr+size);
+            request = new SST::Interfaces::StandardMem::Write(
+                addr, data.size(), data);
+            break;
+        }
+        case gem5::MemCmd::ReadReq: {
+            request = new SST::Interfaces::StandardMem::Read(addr, size);
+            break;
```
powerjg, line 337: I'm not sure I buy this is reasonable. Does this ever happen? If so, what is the result?

ext/sst/sst_responder_subcomponent.hh
```diff
@@ -45,19 +45,26 @@
...
 #include "sst_responder.hh"
 
 class SSTResponderSubComponent: public SST::SubComponent
 {
   private:
-    gem5::OutgoingRequestBridge* responseReceiver;
+    // gem5::OutgoingRequestBridge* responseReceiver;
+    // responseReceiver for this branch is hardcoded to ExternalMemory*.
+    // TODO: We need to make a better design to handle multiple types of
+    // outgoing request classes.
```
powerjg, line 62: This must be done sooner rather than later. We can chat about the best approach.

powerjg (commented):
Overall, this looks very good! Great job cleaning things up.

Let's focus on the experiments now, and we can clean things up futher soon.

Let's focus on writing a paper on this and improving the code so we can release it.

disaggregated_memory/boards/arm_main_board.py
kaustav-goswami: I'll fix the TODO. I copy-pasted the `__metaclass__`. I'll get rid of that.

disaggregated_memory/boards/riscv_main_board.py
kaustav-goswami: So, I am using `self._localMemory` and `self._remoteMemory` for the init phase. I wanted local_memory and remote_memory conventions to exist, similar to the other board, so I kept them as well. I can get rid of these and test the code.

disaggregated_memory/boards/riscv_main_board.py
kaustav-goswami: So `get_mem_ports`' returns a tuple of address range and a port, which is directly used while setting up ruby caches. I did not want to create another function just to get the port or the range. I agree `[0][0]` does look hard-coded.

disaggregated_memory/configs/arm-main.py
kaustav-goswami: I can do that. I'll update this and push this in the next patch.

disaggregated_memory/configs/riscv-main.py
kaustav-goswami: I'll fix this in the next patch.

disaggregated_memory/configs/riscv-main.py
kaustav-goswami: We can package all the experiments as workload objects. I can spend some time to see how we can use it and customize switch CPUs.

disaggregated_memory/configs/riscv-main.py
kaustav-goswami: Sorry about that. This comment was only for the ARM board. I'll also fix the sequence of commands here.

disaggregated_memory/configs/riscv-main.py
kaustav-goswami: But the user might also run gem5-only experiments, right? Or should we completely remove that option from these runscipts?

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: That makes sense. I was thinking about it too. Here is the naming convention that I am thinking of.
`ExternalMemory` -> `AbstractMem`
`ExternalRemoteMemory` -> `AbstractMemorySystem`

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: This is something that we need to discuss. Ranges should be set on the board imo. But we need a way to plug and play ExternalMemory with a range. This is because the remote memory may not be contiguous. So the question is: `Do the board or the memory need to know where the remote memory?`

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: I'll fix this in the next patch.

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: I'll fix this in the next patch.

ext/sst/sst_responder_subcomponent.cc
kaustav-goswami: I'll fix this in the next patch.

ext/sst/sst_responder_subcomponent.cc
kaustav-goswami: I'll fix this in the next patch. The comment is not really relevant but I'll try to word this in a better way.

ext/sst/sst_responder_subcomponent.cc
kaustav-goswami: I'll fix this in the next patch.

ext/sst/sst_responder_subcomponent.cc
kaustav-goswami: This is to mirror functional accesses in SST. I'd still need to decode the gem5 packet so that I can send corresponding timing requests to SST's memory.

ext/sst/sst_responder_subcomponent.hh
kaustav-goswami: I agree. I ran into this issue yesterday so I tried to solve it hastily.

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: I'll fix that in the next patch.

disaggregated_memory/boards/arm_main_board.py
kaustav-goswami: Done!

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: Done!

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: Done!

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: This is completely deleted now! Even the methods are gone.

disaggregated_memory/memories/external_remote_memory_v2.py
kaustav-goswami: done!

disaggregated_memory/configs/riscv-main.py
kaustav-goswami: Done!

disaggregated_memory/configs/riscv-main.py
kaustav-goswami: Done!

disaggregated_memory/configs/arm-main.py
kaustav-goswami: Actually, we should keep these configs separate imo. Otherwise, the user has to compile ALL/gem5. We can discuss more on this.

ext/sst/sst_responder_subcomponent.cc
kaustav-goswami: Done. Also added a small comment.

PR #279, opened 2025-03-24
misc: SRNoC model

kunpai:
This is a model of a statically-scheduled circuit-switched NoC for superconductors.
Modeled off the SRNoC paper by Michelogiannakis et al.


src/network/DataCell.cc
powerjg, line 1: All files need copyright.

powerjg (changes requested):
I see almost no comments. You need to add comments to this code. 

Feel free to use AI to help you, but you have to have comments.

